PR #1, opened 2020-08-12
[BD-03] [TNL-7352] Initial forum structure

xitij2000:
Adds the initial basic structure for the discussions MFE.

**JIRA tickets**: TNL-7352

**Screenshots**: NA. Adds basic component structure that doesn't try to match visuals or functionality. 

**Sandbox URL**: No sandbox is available for this stage. 

**Merge deadline**: None

**Testing instructions**:

NA

**Reviewers**
- [x] @0x9a (initially reviewed here: https://github.com/open-craft/frontend-app-discussions.old/pull/1)
- [ ] @stvstnfrd

PR #2, opened 2020-08-21
[BD-03] [TNL-7411] [TNL-7412] [TNL-7413] Topics, Threads, and Posts UI

pcockwell:
Provides React components and functionality for the Topics, Threads, and Posts interfaces. I would like to add either [MomentJS](https://momentjs.com/) or [timeago.js](https://timeago.org) in order to format datetimes for the individual Posts/Comments/Replies in a `X hours ago` format. Currently, this only directly inserts datetimes.

Please note that some reorganization has occurred:

- Posts, Comments and Replies all belong to Threads
  - There is 1 Post per Thread, and it is the very first user post in the Thread. The Post title is the Thread title
  - There are 0 or more Comments per Thread. These are the top level responses by users within the thread
  - There are 0 or more Replies per Comment. These are responses directly attributed to the parent comment
- Threads belong to Topics
  - There are 0 or more Threads per Topic
- Topics are the same as they have been in the past

**JIRA tickets**:
[BLENDED-557](https://openedx.atlassian.net/browse/BLENDED-557)
[BD-3](https://openedx.atlassian.net/browse/BD-3)
[TNL-7411](https://openedx.atlassian.net/browse/TNL-7411)
[TNL-7412](https://openedx.atlassian.net/browse/TNL-7412)
[TNL-7413](https://openedx.atlassian.net/browse/TNL-7413)

**Dependencies**: 

This PR is directly based off https://github.com/edx/frontend-app-discussions/pull/1

**Screenshots**:

### Topics UI
topics_ui

### Topics UI (filtered to specific category)
topics_subtopic_ui

### All Threads UI

all_threads_ui

### My Threads UI

my_threads_ui

### Threads & Posts
threads_posts_replies_ui

**Testing instructions**:

1. Go to your LMS, login as the `edx` user, and enroll in a course (any course) and create a post in the discussion forum. Repeat with at least 1 other user. Ensure one post contains replies, and there is at least 1 discussion type post, and 1 question type post.
2. Copy the course key for the course you made comments in
3. Launch the frontend-app-discussions application with `npm install && npm start`
4. Navigate to `locahost:2002/discussions/COURSE_KEY/topics`. This is the Topics UI
5. Click on the topic group (with the gray headers) containing the topic you commented in
6. Click on the topic you commented in. This is the Threads UI
7. Click on your thread. This is the Posts/Comments/Replies UI

**Author notes and concerns**:

1. I would like to add either [MomentJS](https://momentjs.com/) or [timeago.js](https://timeago.org) in order to format datetimes for the individual Posts/Comments/Replies in a `X hours ago` format. Currently, this only directly inserts datetimes.
2. There is insufficient API support for this frontend, and as such, some data/information is missing. This UI will likely need to be updated/improved slightly when the API is built/complete.

**Reviewers**
- [ ] @giovannicimolin
- [ ] edX reviewer[s] TBD


src/data/constants.js
```diff
@@ -0,0 +1,64 @@
...
+  BY_COMMENT_COUNT: 'sort_by_comment_count',
+  BY_VOTE_COUNT: 'sort_by_vote_count',
+};
+
+export const ThreadView = {
+  UNREAD: 'unread',
+  UNANSWERED: 'unanswered',
+};
+
+export const MyThreadsFilter = {
```
giovannicimolin, line 23: Any reason why we're changing from `Posts` (`All posts`, `My posts`) to `Threads` in this PR?
Can you adjust it to match the mockups?

src/discussions/topics/TopicsView.jsx
```diff
@@ -0,0 +1,54 @@
...
+          name={topicGroup.name}
+          subtopics={topicGroup.children}
+          key={topicGroup.name}
+        />
+      ),
+    );
+  };
+
+  return (
+    <div className="discussion-topics d-flex flex-column card">
```
giovannicimolin, line 39: This is missing the filter beside the search bar (https://www.figma.com/proto/LVFnO9oRHYiamoicZuXDAC/Discussion-page?node-id=238%3A1676&scaling=min-zoom)

src/data/messages.js
```diff
@@ -0,0 +1,41 @@
...
+
+const messages = defineMessages({
+  all_threads: {
+    id: 'discussions.navigation-bar.filter.all-threads',
+    defaultMessage: 'All threads',
+    description: 'Option in dropdown to filter to all threads',
+  },
+  all_topics: {
+    id: 'discussions.navigation-bar.filter.all-topics',
+    defaultMessage: 'All topics',
```
giovannicimolin, line 11: This is just `Topics` in the mockups.

src/discussions/posts/post/Post.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import postShape from './post-shape';
+import PostIcons from '../post-icons/PostIcons';
+
+function Post({ post }) {
+  return (
+    <div className="discussion-post d-flex flex-column">
```
giovannicimolin, line 7: This is the threads UI, and it doesn't look close to the one in the mockups.
There's borders around the components, the sizes look different and the post editor doesn't show up.

The buttons on the post UI don't work.

**Mockup:**
image

**Implementation:**
image

src/discussions/navigation-bar/NavigationBar.jsx
```diff
@@ -0,0 +1,62 @@
...
+import React from 'react';
+import { useParams } from 'react-router';
+import { NavLink } from 'react-router-dom';
+import { Routes } from '../../data/constants';
+import messages from '../../data/messages';
+
+function NavigationBar({ intl }) {
+  const { courseId } = useParams();
+  return (
+    <div className="navigation-bar d-flex flex-row my-2">
```
giovannicimolin, line 13: After clicking on a thread, the navigation buttons loose the indication of which page the user is in.
image

src/discussions/threads/ThreadsView.jsx
```diff
@@ -0,0 +1,26 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import ThreadFilterBar from './thread-filter-bar/ThreadFilterBar';
+import Thread, { threadShape } from './thread/Thread';
+
+function ThreadsView({ threads, filterSelfThreads }) {
+  return (
+    <div className="discussion-threads d-flex flex-column card">
```
giovannicimolin, line 8: This view is missing the breadcrumbs with the navigation structure present in the mockups.
**Mockups:**
image

src/discussions/threads/thread/Thread.jsx
```diff
@@ -0,0 +1,99 @@
...
+} from '@fortawesome/free-solid-svg-icons';
+
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import PropTypes from 'prop-types';
+import React from 'react';
+import { Link } from 'react-router-dom';
+import { Routes } from '../../../data/constants';
+import messages from './messages';
+
+function Thread({ thread, intl }) {
```
giovannicimolin, line 14: This component doesn't match the mockups.
- There's no star at the bottom of the card
- The [A] should be the user's avatar (or a placeholder with dummy image)
- The black dot should be the online/offline status (looks off right now)
- There's no indication that a post is pinned (pin in the icon on mockups)

**Mockup:**
image

**Implemented:**
image

src/discussions/threads/ThreadsView.jsx
giovannicimolin: https://edx.github.io/paragon/components/breadcrumbs

src/components/selectable-dropdown/SelectableDropdown.jsx
```diff
@@ -0,0 +1,55 @@
+import { Dropdown } from '@edx/paragon';
+import PropTypes from 'prop-types';
+import React, { useState } from 'react';
+
+function SelectableDropdown({
```
giovannicimolin, line 5: In Paragon (https://edx.github.io/paragon/components/dropdown) is just a pass-through of the React Bootstrap component.

Using https://react-bootstrap.github.io/components/dropdowns/#custom-dropdown-components would be a better match for this ui.

src/discussions/navigation-bar/NavigationBar.jsx
giovannicimolin: These should implemented using the Nav component from Paragon:
https://edx.github.io/paragon/components/navs.

This was already written down in mockups as `Developer Notes`. Is there any issue in using the recommended components?
https://www.figma.com/proto/LVFnO9oRHYiamoicZuXDAC/Discussion-page?node-id=298%3A0&scaling=min-zoom
image

Makefile
```diff
@@ -0,0 +1,54 @@
+transifex_resource = frontend-template-application
```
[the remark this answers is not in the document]
pcockwell: This is built on top of #1

src/data/constants.js
pcockwell: I changed it because the terminology used in the mockups do not match the terminology used in the current edx-platfor API, and in general, the definitions of the words in the English language.

The [edx-platform API](https://github.com/edx/edx-platform/blob/master/lms/djangoapps/discussion/rest_api/views.py#L145) uses the term 'Thread' to describe what the mockups label as a 'Post'.

The definition of the term 'Post' on [Cambridge Dictionary](https://dictionary.cambridge.org/dictionary/english/post) is:
```
something such as a message or picture that you publish on a website or using social media
```

Conversely, the definition of 'Thread' on [Cambridge Dictionary](https://dictionary.cambridge.org/dictionary/english/thread) is:
```
The thread of a book, discussion, speech, etc. is its story or the way that it develops, one part connecting with another
```

This can easily be changed back, but I believe this terminology of using 'Thread' as a collection of messages from users, the first of which is a 'Post', and subsequent ones being either a 'Comment' or a 'Reply' is better. Please advise if this needs to be changed.

src/discussions/navigation-bar/NavigationBar.jsx
pcockwell: @giovannicimolin

We are currently using edx/paragon version `10.0.1`, which does not have the `Nav` component. Please see [here|https://github.com/edx/paragon/tree/v10.0.1/src]

src/components/selectable-dropdown/SelectableDropdown.jsx
pcockwell: @giovannicimolin

I'm not quite sure I follow you... The `SelectableDropdown` component uses the React Bootstrap `Dropdown`, `Dropdown.Toggle`, `Dropdown.Menu` and `Dropdown.Item` components that are passed through. I tend to think the UI for the dropdown elements is rendered well (when comparing to the mockups). What exactly do think doesn't look right? How should the custom dropdown components be used to improve the look?

openedx.yaml
```diff
@@ -0,0 +1,12 @@
...
+
+nick: tmpa
+oeps: {}
+owner: edx/arch-team
+openedx-release:
+  # The openedx-release key is described in OEP-10:
+  #   https://open-edx-proposals.readthedocs.io/en/latest/oep-0010-proc-openedx-releases.html
+  # The FAQ might also be helpful: https://openedx.atlassian.net/wiki/spaces/COMM/pages/1331268879/Open+edX+Release+FAQ
+  maybe: true   # Delete this "maybe" line when you have decided about Open edX inclusion.
+  ref: master
```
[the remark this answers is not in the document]
xitij2000: I don't know enough to fill this in. Waiting for a review of my own PR for that.

src/discussions/navigation-bar/NavigationBar.jsx
xitij2000: > We are currently using edx/paragon version `10.0.1`, which does not have the `Nav` component. Please see [here](https://github.com/edx/paragon/tree/v10.0.1/src)

@pcockwell Yes it does: https://github.com/edx/paragon/blob/v10.0.1/src/index.js#L106
If you visit the [link mentioned on the notes page](https://edx.github.io/paragon/components/navs) you will learn that it is a pass-through component to React-Bootstrap.

src/discussions/threads/ThreadsView.jsx
xitij2000: @giovannicimolin I think it's OK for that to not be there, since the focus here is on the internal UI. The overall UI was meant to be a different task.

src/discussions/threads/thread/Thread.jsx
pcockwell: The star at the bottom of the card is something that was introduced by Kshitij in #1. @xitij2000, should the star be removed or is that something we want in the Thread UI?

I've replaced the `[A]` with a round image that currently sources a random flower image.

The black dot is labelled in the code (from the previous PR) as the read/unread status. @xitij2000 could you provide context as to whether the dot is intended to be read/unread status, or online/offline status for the user. If it is supposed to be read/unread status, under which circumstances should the circle be visible, and what color should it be in those circumstances?

There is in fact code for the pinned status of a post. It used to show up on the right side, but will now be displayed at the bottom alongside the star and comment count.

src/components/selectable-dropdown/SelectableDropdown.jsx
xitij2000: I think this component can probably be lifted up to Paragon if it's used enough, for now we can just use it as-is.

src/discussions/posts/post/Post.jsx
xitij2000: @giovannicimolin This is mostly about the UI, it's OK if the buttons don't work yet, and the post editor work is separate.

xitij2000:
@giovannicimolin I think fixing it via SCSS should be avoided, but wherever you can get better, closer styling via bootstrap, that should be attempted. 

Note: currently the new post UI is in development, and the filter bar, breadcrumbs etc were not in scope of these tickets

src/discussions/threads/thread/Thread.jsx
xitij2000: @pcockwell As I mentioned earlier I implemented the original skeletal code based on the older mockups. The basic sturcure is roughly the same, but if in doubt follow the mockups I shared, not my implementaiton.

src/discussions/posts/post/Post.jsx
pcockwell: @giovannicimolin @xitij2000

The issue with the mockups is that they do not indicate how a post with replies should look. What does the full thread look like when there are more than the original post.  If the design I've implemented using the `Card` UI that was used for the Threads component is not what we want, I am happy to rework this. The `Card` UI was the simplest way to delineate the differences between each individual post/comment/reply.

I'll look into adding the post editor. Do we have a component that we use for this? I assume there is a specific HTML/JSX/React snippet I should be using for this and that I don't need to reimplement all of it.

As for the buttons, I was under the impression that the extent of the functionality I was to implement for this was the interactions that send a user to the existing components (Topic, Thread, Post, Comment etc) and that any additional functionality would be added to the MFE separately.

src/discussions/threads/thread/Thread.jsx
xitij2000: > The black dot is labelled in the code (from the previous PR) as the read/unread status. @xitij2000 could you provide context as to whether the dot is intended to be read/unread status, or online/offline status for the user. If it is supposed to be read/unread status, under which circumstances should the circle be visible, and what color should it be in those circumstances?
It is the unread status. It shows up if unread, and doesn't show up if read. There is no other colour.

pcockwell:
@xitij2000 

I've actually gone and added the breadcrumbs already and fixed the topic filter bar. By 'new post UI', do you mean the post editor?

src/discussions/posts/post/Post.jsx
xitij2000: > The issue with the mockups is that they do not indicate how a post with replies should look.

You can implement it in a straight line for now. 

> I'll look into adding the post editor. Do we have a component that we use for this? I assume there is a specific HTML/JSX/React snippet I should be using for this and that I don't need to reimplement all of it.

Do not do that. The scope of this ticket is limited to the things I mentioned on the ticket. This ticket is already going way beyond that scope. 

> As for the buttons, I was under the impression that the extent of the functionality I was to implement for this was the interactions that send a user to the existing components (Topic, Thread, Post, Comment etc) and that any additional functionality would be added to the MFE separately.

Correct, but it's useful to add stub code to allow exploring all the UI.

xitij2000:
> By 'new post UI', do you mean the post editor?

Yes.

src/discussions/posts/post/Post.jsx
pcockwell: @xitij2000

> > The issue with the mockups is that they do not indicate how a post with replies should look.
>
> You can implement it in a straight line for now.

So, to confirm, the desired result is to remove the `Card` UI, and then simply render all comments in a straight line one after another with no separator? Also, how should the Replies be rendered since the original edx-platform implementation for discussions shows them as an inline thread immediately after the comment they are replying to?

src/discussions/navigation-bar/NavigationBar.jsx
pcockwell: @xitij2000 @giovannicimolin

I've been playing around with the `Nav` component. I don't particularly think the user experience on it is very good. When you click an element in the `Nav`, it actually does a full page reload as it navigates to that URL, where as the `NavLink` code I had previously, uses react routing to send the user to that page. Further, without using SCSS I cannot find a way to style the `active` element, and the current classes have it simply be a black text item (as opposed to the default blue) and does not have any background color.

My recommendation is to remain with the current `NavLink` implementation, though if desired I can push a change that will use the `Nav` component so you can compare the results.

xitij2000: That happens because the Nav component uses the `a` tag by default. However, it includes a mechanism to provide your own link component via the `as` attribute. For instance you could do:

```jsx
<Nav.Link as={NavLink} to={url}>Text</Nav.Link>
```

src/discussions/posts/post/Post.jsx
xitij2000: The desired result is to get the UI to look as close to the mockup as possible. If using Cards helps, then you use them. If not, you don't. In this case they seem to do more harm than good. 

This is a first-pass implementation, and the idea is to get a very close representation of what a single comment/post looks like since we have that mockup.

src/data/messages.js
```diff
@@ -0,0 +1,45 @@
...
+    id: 'discussions.navigation-bar.sort.comment-count',
+    defaultMessage: 'Most activity',
+  },
+  sort_by_vote_count: {
+    id: 'discussions.navigation-bar.sort.vote-count',
+    defaultMessage: 'Most votes',
+  },
+});
+
+export default messages;
```
xitij2000, line 45: Why is this in the data folder? Move it to the navigation bar component.

pcockwell: Aside from `all_topics`, `all_threads` and `my_threads`, none of these are used by the navigation bar component, and in some cases, these are used in multiple places elsewhere, which is why I moved it into `data`. I'll see if I can more accurately rename the `id` for some of these.

src/data/constants.js
giovannicimolin: I still think we should follow the mockups, but this is something I need more opinions on. @xitij2000 can you pitch in here?
Also, can you ask this upstream somewhere? Make sure to ping me and Kshitij there.

src/components/selectable-dropdown/SelectableDropdown.jsx
giovannicimolin: I don't think it needs to be lifted up to Paragon, as it provides enough flexibility to match the mockups as is. The link I've posted above has an example that is pretty close to the desired result (it's a combination of an `a` tag + Dropdown menu and only uses bootstrap classes).

src/data/constants.js
xitij2000: > I changed it because the terminology used in the mockups do not match the terminology used in the current edx-platfor API, and in general, the definitions of the words in the English language.

We definitely need to follow the mockups. The terms used in the API are mostly inconsequential since:
- End-users will not see the API
- We are doing a redesign of the UI and potentially the APIs as well

> The [edx-platform API](https://github.com/edx/edx-platform/blob/master/lms/djangoapps/discussion/rest_api/views.py#L145) uses the term 'Thread' to describe what the mockups label as a 'Post'.

Yes, the API has topics, threads and comments. The UI doesn't use the same terms, we're following the mockup, not the existing API. What you should look at for reference is the existing UI, which uses the same terms.

This is the topics UI:
Screenshot_2020-09-02 Discussion DemoX OpenCraft - Demo Instance

This is the posts UI. Note that the term here is "show all posts":
Screenshot_2020-09-02 Discussion DemoX OpenCraft - Demo Instance(1)

> This can easily be changed back, but I believe this terminology of using 'Thread' as a collection of messages from users, the first of which is a 'Post', and subsequent ones being either a 'Comment' or a 'Reply' is better. Please advise if this needs to be changed.

I understand, but that discussion is not in scope here. The existing UI and new UI are consistent in how they name things and we need to maintain that.

src/data/messages.js
xitij2000: If strings are used twice, define them twice. If they are used in two different contexts they need to have two different IDs.

src/discussions/posts/post/post-shape.js
```diff
@@ -0,0 +1,10 @@
+import PropTypes from 'prop-types';
+
+const postShape = PropTypes.shape({
+  posted_on: PropTypes.string,
+  abuse_flagged: PropTypes.bool,
+  rendered_body: PropTypes.string,
+  author: PropTypes.string,
+});
+
+export default postShape;
```
xitij2000, line 10: Put the shape in the same file as the component, it can be imported from there. Don't make a separate file.

src/discussions/posts/reply/Reply.jsx
```diff
@@ -0,0 +1,33 @@
...
+              <div className="comment-body d-flex" dangerouslySetInnerHTML={{ __html: reply.rendered_body }} />
+            </div>
+            <PostIcons post={reply} />
+          </div>
+          <div className="visibility-comment d-flex small text-gray-300">
+            This post is visible to everyone
+          </div>
+        </div>
+      </div>
+    </div>
```
xitij2000, line 25: Are so many layered divs really needed?

src/discussions/threads/thread-filter-bar/ThreadFilterBar.jsx
```diff
@@ -0,0 +1,86 @@
...
+
+  useEffect(() => {
+    setThreadFilter(defaultThreadFilter);
+  }, [filterSelfThreads]);
+
+  const threadsStatusFilterOptions = buildIntlSelectionList(ThreadsStatusFilter, intl, messages);
+  const threadOrderingOptions = buildIntlSelectionList(ThreadOrdering, intl, messages);
+
+  const [currentThreadStatus, setThreadStatus] = useState(ThreadsStatusFilter.ALL);
+  const [currentThreadOrder, setThreadOrder] = useState(ThreadOrdering.BY_LAST_ACTIVITY);
```
xitij2000, line 31: This should be a dumb component that doesn't store any state. What is the point of storing the current filter internally in the component when it needs to affect the data fetched and the whole UI?

src/discussions/topics/topic-group/topic/Topic.jsx
```diff
@@ -1,58 +1,72 @@
...
 }
 
 export const topicShape = {
   name: PropTypes.string,
   id: PropTypes.string,
-  topics: PropTypes.array.isRequired, // eslint-disable-line react/forbid-prop-types
+  subtopics: PropTypes.array,
+  questions: PropTypes.arrayOf(PropTypes.object),
+  discussions: PropTypes.arrayOf(PropTypes.object),
+  flags: PropTypes.arrayOf(PropTypes.object),
```
xitij2000, line 60: This approach is incorrect. This is now how the API will return the data. These will just be numbers.

src/discussions/topics/topic-search-bar/messages.js
```diff
@@ -0,0 +1,11 @@
+import { defineMessages } from '@edx/frontend-platform/i18n';
+import globalMessages from '../../../data/messages';
+
+const messages = defineMessages({
+  sorted_by: globalMessages.sorted_by,
+  sort_by_last_activity: globalMessages.sort_by_last_activity,
+  sort_by_comment_count: globalMessages.sort_by_comment_count,
+  sort_by_course_structure: globalMessages.sort_by_course_structure,
+});
```
xitij2000, line 9: Don't do this. Define the messages separately.

pcockwell:
@giovannicimolin @xitij2000 

I believe I've addressed all comments in this PR thus far except for @xitij2000's comment about the [`PostFilterBar`](https://github.com/edx/frontend-app-discussions/pull/2#discussion_r481729632) (which I will get to ASAP). Could you guys do another round of review when possible?

src/discussions/threads/thread-filter-bar/ThreadFilterBar.jsx
pcockwell: @xitij2000

Initially when I wrote this component, I fully expected that it would need a rework to properly store the state so that it can be used by API calls. I was not, and still am not really sure how to approach refactoring this component such that it uses the state appropriately. This is the component that sets the values for filtering/ordering of posts, so it needs to have some mechanism to set those values, whether that be as local state, or in the redux state. I can definitely make it such that it retrieves the existing filtering/ordering from the state, but in order for this component to work properly, I would also have to implement some amount of `slices` or `thunks` code to set/modify that state.

I can try some things out and see what I come up with, but (as you've pointed out) this PR is already somewhat bloated and I feel like this would likely add more bloat. Do you want me to make an attempt at implementation for this, or save it for a future PR?

xitij2000: > it needs to have some mechanism to set those values, whether that be as local state, or in the redux state.

I think it only makes sense to store it in redux since storing it locally makes it inaccessible to the rest of the app. So adding code to store it locally doesn't make sense. 

> I can definitely make it such that it retrieves the existing filtering/ordering from the state, but in order for this component to work properly, I would also have to implement some amount of `slices` or `thunks` code to set/modify that state.

The purpose of this task is to implement the UI designs, not to hook up everything properly, As I mentioned earlier there are, and will be other tasks to do that kind of hooking up. 

i.e. I am asking you to remove the existing features, not develop them more thoroughly. You can just plop the component in there as a bit of non-functional UI.

pcockwell: @xitij2000

I've removed the state usage from the `PostFilterBar`. It still takes into account whether it should show `My Posts` or `All Posts` (and the associated alternative options), but does not update the filter bar, nor do anything else when you select an option from the dropdown.

giovannicimolin:
@pcockwell  Addressing your concerns:

> I would like to add either MomentJS or timeago.js in order to format datetimes for the individual Posts/Comments/Replies in a X hours ago format. Currently, this only directly inserts datetimes.

You can use MomentJS, as it's already used in other frontend repos (see https://github.com/edx/frontend-app-publisher/blob/master/package.json#L40).

>There is insufficient API support for this frontend, and as such, some data/information is missing. This UI will likely need to be updated/improved slightly when the API is built/complete.

There's no problem with that, the goal of this ticket is to implement a base to work on.

src/discussions/posts/post/Post.jsx
```diff
@@ -0,0 +1,111 @@
...
+      </div>
+      <div className="d-flex flex-column">
+        <div className="d-flex post-header">
+          <div className="d-flex user-avatar col-2 m-1 px-0">
+            <Image
+              className="w-75"
+              src="https://source.unsplash.com/100x100/?nature,flower"
+              roundedCircle
+            />
+          </div>
```
giovannicimolin, line 36:
```suggestion
          <Image
            src="https://source.unsplash.com/100x100/?nature,flower"
            roundedCircle
            className="col-1 my-auto p-0"
          />
```

src/discussions/posts/post/Post.jsx
```diff
@@ -0,0 +1,111 @@
...
+      className="discussion-post d-flex list-group-item p-2 text-decoration-none text-gray-900"
+      data-post-id={post.id}
+      to={
+        Routes.COMMENTS.PATH.replace(':courseId', post.course_id)
+          .replace(':topicId', post.topic_id)
+          .replace(':postId', post.id)
+      }
+    >
+      <div className="d-flex post-unread-status m-1">
+        { post.read || <FontAwesomeIcon icon={faCircle} className="text-success-300" /> }
```
giovannicimolin, line 26:
```suggestion
      <div className="post-unread-status mx-1">
        { post.read || <FontAwesomeIcon icon={faCircle} className="text-success-300" size="xs"/> }
```

src/discussions/posts/post/Post.jsx
```diff
@@ -0,0 +1,111 @@
...
+            </div>
+          </div>
+        </div>
+        <div className="d-flex my-1">
+          { post.raw_body }
+        </div>
+        <div className="d-flex">
+          { post.pinned && (
+            <span className="badge mx-1">
+              <FontAwesomeIcon icon={faThumbtack} />
```
giovannicimolin, line 68: These and the icons below can be `size=lg`

src/discussions/posts/post/Post.jsx
```diff
@@ -0,0 +1,111 @@
...
+                    ? intl.formatMessage(messages.last_response, { time: post.updated_at })
+                    : intl.formatMessage(messages.posted_on, { time: post.created_at })
+                }
+              </div>
+            </div>
+          </div>
+        </div>
+        <div className="d-flex my-1">
+          { post.raw_body }
+        </div>
```
giovannicimolin, line 64:
```suggestion
        <small className="d-flex my-2">
          { post.raw_body }
        </small>
```

giovannicimolin (changes requested):
A few style suggestions to better match the mockups.

src/discussions/comments/comment-icons/CommentIcons.jsx
```diff
@@ -0,0 +1,32 @@
...
+  return (
+    <div className="d-flex flex-column icons">
+      {/* Only show the star if the comment has a following attribute, indicating it can be followed */}
+      { following !== undefined && (
+        following
+          ? <FontAwesomeIcon icon={faSolidStar} />
+          : <FontAwesomeIcon icon={faEmptyStar} />
+      )}
+      { abuseFlagged && <FontAwesomeIcon className="my-1 mx-auto" icon={faFlag} /> }
+      <FontAwesomeIcon className="my-1 mx-auto" icon={faEllipsisV} />
```
giovannicimolin, line 17: These elements should be buttons.
```suggestion
      {/* Only show the star if the comment has a following attribute, indicating it can be followed */}
      { following !== undefined && (
        following
          ? (
            <Button variant="link" className="p-0" size="xs">
              <FontAwesomeIcon icon={faSolidStar} />
            </Button>
          ) : (
            <Button variant="link" className="p-0" size="xs">
              <FontAwesomeIcon icon={faEmptyStar} />
            </Button>
          )
      )}
      { abuseFlagged && (
        <Button variant="link" className="p-0" size="xs">
          <FontAwesomeIcon icon={faFlag} />
        </Button>
      ) }
      <Button variant="link" className="p-0" size="xs">
        <FontAwesomeIcon icon={faEllipsisV} />
      </Button>
```

src/discussions/comments/comment/Comment.jsx
```diff
@@ -0,0 +1,40 @@
...
+
+function Comment({ comment }) {
+  return (
+    <div className="discussion-comment d-flex flex-column">
+      <div className="header d-flex flex-row">
+        <div className="d-flex flex-column flex-fill">
+          <div className="title h4">
+            { comment.title }
+          </div>
+          <div className="status small">
```
giovannicimolin, line 13: Good to go as is, but I feel we're lacking an even smaller font size for a few things (CC @xitij2000).

src/discussions/posts/post/Post.jsx
```diff
@@ -0,0 +1,111 @@
...
+import React from 'react';
+import { Link } from 'react-router-dom';
+import { Image } from '@edx/paragon';
+import { Routes } from '../../../data/constants';
+import messages from './messages';
+
+function Post({ post, intl }) {
+  return (
+    <Link
+      className="discussion-post d-flex list-group-item p-2 text-decoration-none text-gray-900"
```
giovannicimolin, line 17:
```suggestion
      className="discussion-post d-flex p-3 text-decoration-none text-gray-900"
```

src/discussions/posts/post/Post.jsx
```diff
@@ -0,0 +1,111 @@
...
+                {
+                  post.updated_at
+                    ? intl.formatMessage(messages.last_response, { time: post.updated_at })
+                    : intl.formatMessage(messages.posted_on, { time: post.created_at })
+                }
+              </div>
+            </div>
+          </div>
+        </div>
+        <div className="d-flex my-1">
```
giovannicimolin, line 62:
```suggestion
        <div className="d-flex my-1 small">
```

src/discussions/posts/post/Post.jsx
```diff
@@ -0,0 +1,111 @@
...
+import PropTypes from 'prop-types';
+import React from 'react';
+import { Link } from 'react-router-dom';
+import { Image } from '@edx/paragon';
+import { Routes } from '../../../data/constants';
+import messages from './messages';
+
+function Post({ post, intl }) {
+  return (
+    <Link
```
giovannicimolin, line 16: I this isn't exactly in the scope here, but clicking on a post, no matter in which page, will make the page navigate to the posts view (or at least change the `Nav` to that).

nav_issue

This is likely due to the page structure and will need to be reworked in a future iteration.

CC @xitij2000

giovannicimolin (changes requested):
@pcockwell  @xitij2000  The scope of this is kinda too big, and there's too many changes to review, so I might miss a few things.

After reading the code and testing out a few things locally, I feel like there's a lot of places over-using bootstrap classes and that the implementation can be simplified by a better use of those.

I haven't reviewed the structure around the views (I considering it prototype boilerplate code), but I can see how the general structure needs to be improved to remove UI quirks (like the one mentioned on a comment below).

src/discussions/topics/topic-group/topic/Topic.jsx
```diff
@@ -0,0 +1,72 @@
...
+  return (
+    <Link
+      className="discussion-topic d-flex flex-column list-group-item text-gray-900 text-decoration-none p-2"
+      data-topic-id={id}
+      to={
+        Routes.POSTS.PATH.replace(':courseId', courseId)
+          .replace(':topicId', id)
+      }
+    >
+      <div className="topic-name h6">
```
giovannicimolin, line 30: This component can be improved a lot to match the mockups only using bootstrap classes.
Items to be improved: 

- Font sizes, color and decoration 
- Icons sizes and color
- Remove border around each item (`list-group-item`)
- Improve spacing

Ref: https://www.figma.com/proto/LVFnO9oRHYiamoicZuXDAC/Discussion-page?node-id=238%3A1676&scaling=min-zoom

pcockwell: @giovannicimolin

> Remove border around each item (list-group-item)

While the styling for the borders might have been stronger than in the mockup, I do see borders in between each item (though they are faint). Should I be removing them entirely?

xitij2000:
> You can use MomentJS, as it's already used in other frontend repos (see https://github.com/edx/frontend-app-publisher/blob/master/package.json#L40).

We discussed this on the TNL ticket and the conclusion was that we should use timeago instead of momentjs since momentjs is too heavy.

src/discussions/posts/post/Post.jsx
xitij2000: The image should not be hardcoded, it should be coming from the parent component. It can be hardcoded at some point, but not here.

src/discussions/comments/comment/Comment.jsx
xitij2000: Noted, we can suggest a new size at the paragon level.

src/discussions/posts/post/Post.jsx
xitij2000: I think this is OK to leave for another PR. This is because there isn't an unambiguous way to determine which view we are on from the current URL. Chances are this will need either a change in the URL structure or we can save this info to redux.

pcockwell:
> > You can use MomentJS, as it's already used in other frontend repos (see https://github.com/edx/frontend-app-publisher/blob/master/package.json#L40).
> 
> We discussed this on the TNL ticket and the conclusion was that we should use timeago instead of momentjs since momentjs is too heavy.

Correct, which is why I installed `timeago.js`

xitij2000:
> Correct, which is why I installed `timeago.js`

Sorry, that was directed towards @giovannicimolin as the reviewer.

src/discussions/topics/topic-group/topic/Topic.jsx
giovannicimolin: >While the styling for the borders might have been stronger than in the mockup, I do see borders in between each item (though they are faint). 

I guess using my monitor in a warm setting didn't help in getting me to see it. I've switched to `game` mode the lines magically appeared.

>Should I be removing them entirely?

You should keep them.

src/discussions/posts/post/Post.jsx
giovannicimolin: :+1:

giovannicimolin (commented):
@xitij2000 @pcockwell I've read and tested the code as much as I could, but I don't think is reasonable for us to push this amount of code in a single go to an upstream repository (at least I don't feel comfortable approving a PR this size).

A few reasons:
1. A big amount of this code is prototype level, boilerplate code.
2. There's no tests.
3. This mixes up 3 tasks: topic, thread and posts UI implementation.

We should split this up into multiple, manageable sized PRs, and completely separate UI from the 3 goal tasks from boilerplate code (even rewriting git history is a challenge with the current size of the changes).

With a PR with the current size, it's pretty easy to let things slip through (as I've seen happen in the LTI project with 10x less lines).

Can you suggest an approach/breakdown we can follow to make this easier to review? 
Also, can we bring the edX technical contact into this discussion so we're all aligned in the merge approach?

PR #8, opened 2021-06-08
feat: Add topic listing views, post listing view, comment views etc. [BD-38] [TNL-7413] [BB-2856]

xitij2000:
This PR adds a major chunk of UI code for the discussions MFE. 
It adds the following features:
1. Topic listing
2. Topic filtering (client-side)
3. Subtopic browsing (clicking on a topic category header to see only subtopics in a particular category).
4. Browsing posts in a topic
5. Filtering posts by unread and following but *not* flagged
6. Sorting posts by recent activity, most activity and most votes
5. Liking/Unliking, Following/unfollowing a post
6. Listing comments in a post.
7. Listing replies to a comment in a post. 

Known issues to be resolved in separate PR:
1. Clicking on a post in all posts will cause the highlight to move to "All topics"
2. If you are viewing a post  and change a filter such that that post is no longer visible, then comments for that post will disappear and a spinner will keep spinning. 

This is mostly based on Patrick's PR: https://github.com/edx/frontend-app-discussions/pull/2

**JIRA tickets**: TNL-7413

**Screenshots**:

https://user-images.githubusercontent.com/118837/121207276-b001e400-c868-11eb-9694-da7d195ae819.mp4

**Sandbox URL**: https://forum.discussions.sandbox.opencraft.hosting/discussions/course-v1:edX+DemoX+Demo_Course/topics

**Reviewers**
- [ ] 
- [ ] edX reviewer[s] TBD


edx-webhook:
Thanks for the pull request, @xitij2000! I've created [BLENDED-881](https://openedx.atlassian.net/browse/BLENDED-881) to keep track of it in Jira.


public/index.html
```diff
@@ -1,10 +1,10 @@
 <!doctype html>
 <html lang="en-us">
   <head>
-      <title>Discussions | edX</title>
+      <title>Discussions | <%= process.env.SITE_NAME %></title>
```
giovannicimolin, line 4: Not really a comment about this PR - I see the same pattern on all frontend apps. Is it possible to translate this string?

src/discussions/topics/topic-search-bar/TopicSearchBar.jsx
```diff
@@ -0,0 +1,39 @@
...
+import messages from './messages';
+
+function TopicSearchBar({ intl }) {
+  const dispatch = useDispatch();
+  const topicOrderingOptions = buildIntlSelectionList(TopicOrdering, intl, messages);
+  return (
+    <div className="d-flex flex-row p-1 align-items-center">
+      <SearchField
+        className="flex-fill m-1 border-0"
+        placeholder="Find a topic"
```
giovannicimolin, line 20: `"Find a topic"` needs to be translatable.

public/index.html
xitij2000: I don't think so! However we can switch to using react-helmet, which will allow updating this dynamically per page. That can then use translations.

.nvmrc
```diff
@@ -0,0 +1,2 @@
+12.10.0
```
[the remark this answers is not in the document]
xitij2000: Yup. I copied it over from here: https://github.com/edx/frontend-template-application/blob/master/.nvmrc

src/discussions/comments/data/slices.js
```diff
@@ -3,14 +3,28 @@ import { createSlice } from '@reduxjs/toolkit';
...
+        if (!threads[comment.thread_id]) {
+          threads[comment.thread_id] = [];
+        }
+
+        if (!threads[comment.thread_id].includes(comment.id)) {
+          threads[comment.thread_id].push(comment.id);
+        }
       }
       comments[comment.id] = comment;
     },
```
xitij2000, line 30: This code is reformatting the API response into something that's easier for the UI to look up. 

So, when we call an API and get a list of comments, here is what we do:
- We add it to the comment map using its id. So if we ever need to look up the comment by its id, we can just do state.comments[commentId] to get the comment.
- If the comment has a parent id, it is a comment on a comment, i.e. a response. So, we add it to the list of replies to a comment. For this, we maintain a dictionary of comment ids, to the responses to those comments. 
- If the comment doesn't have a parent id it is a comment on the main thread. So, we add it to a list of the comments in the that thread as above. 

So we take the following three comments:
- {id: "comment1", thread_id: "tid1", parent_id: null, ...}
- {id: "response1", thread_id: "tid1", parent_id: "comment1", ...}
- {id: "comment2", thread_id: "tid1", parent_id: null, ...}

And create the following:
```javascript 
    comments = { 
        comment1: {id: "comment1", thread_id: "tid1", parent_id: null, ...},
        response1: {id: "response1", thread_id: "tid1", parent_id: "comment1", ...}
        comment2: {id: "comment2", thread_id: "tid1", parent_id: null, ...}
    }
    threadCommentMap = {
        tid1: ["comment1", "comment2"]  //note doesn't list response1
    }
    commentResponseMap = {
        comment1: ["response1"]
    }
```

src/discussions/comments/data/slices.js
```diff
@@ -115,14 +132,14 @@ const commentsSlice = createSlice({
...
+  fetchCommentResponsesRequest,
+  fetchCommentResponsesSuccess,
   fetchCommentsDenied,
   fetchCommentsFailed,
   fetchCommentsRequest,
   fetchCommentsSuccess,
-  fetchCommentSuccess,
   postCommentDenied,
   postCommentFailed,
   postCommentRequest,
```
xitij2000, line 145: The rest are mostly code changes due to renaming fetchComment* to the more accurate fetchCommentaResponses*

src/discussions/discussions-home/DiscussionsHome.jsx
```diff
@@ -1,30 +1,38 @@
...
+        </Switch>
+      </div>
+      <div className="d-flex w-50 pl-1 flex-column">
+        <PostActionsBar />
+        <Route path={Routes.COMMENTS.PATH}>
+          <CommentsView />
+        </Route>
       </div>
     </main>
   );
```
xitij2000, line 38: Removed unnecessary nested div.

src/discussions/posts/post/Post.jsx
```diff
@@ -1,76 +1,164 @@
...
+          />
+        </OverlayTrigger>
+      </div>
+    </div>
+  );
+}
+
 export const postShape = PropTypes.shape({
   abuse_flagged: PropTypes.bool,
   author: PropTypes.string,
```
xitij2000, line 164: Refactored this into smaller sucomponents

src/index.scss
```diff
@@ -1,6 +1,10 @@
...
+@import "@edx/brand/paragon/fonts.scss";
+@import "@edx/brand/paragon/variables.scss";
+@import "@edx/paragon/scss/core/core.scss";
+@import "@edx/brand/paragon/overrides.scss";
 
 @import "~@edx/frontend-component-header/dist/index";
 @import "~@edx/frontend-component-footer/dist/footer";
+
+$fa-font-path: "~font-awesome/fonts";
+@import "~font-awesome/scss/font-awesome";
```
xitij2000, line 10: Just refactored a bit to match new template.

xitij2000 (commented):
I've made a number of general changes. 
- I've cleaned up the code a bit to remove too many nested divs
- I've changed the classes a bit to match themockups better. 
- I've removed containers, since with hooks they don't make as much sense.

package.json
```diff
@@ -15,6 +15,7 @@
     "i18n_extract": "BABEL_ENV=i18n fedx-scripts babel src --quiet > /dev/null",
     "is-es5": "es-check es5 ./dist/*.js",
     "lint": "fedx-scripts eslint --ext .js --ext .jsx .",
+    "lint:fix": "fedx-scripts eslint --ext .js --ext .jsx . --fix",
```
kewne, line 18: Nit: this can be achieved through `npm run lint -- --fix`

xitij2000: Yup. I'm just trying to reduce the diff with the template application on which this is based: https://github.com/edx/frontend-template-application/blob/41e8fd4043a9434de20ff7ec2df0452051ebefae/package.json#L18

src/discussions/topics/topic-group/TopicGroup.jsx
```diff
@@ -1,24 +1,68 @@
+import PropTypes from 'prop-types';
 import React from 'react';
+import { useParams } from 'react-router';
+import { Link } from 'react-router-dom';
+import { Routes } from '../../../data/constants';
 import Topic, { topicShape } from './topic/Topic';
 
-function TopicGroup({ id, name, topics }) {
+function TopicGroup({
```
kewne, line 8: IMO it would be better to introduce a `FilteredTopicGroups` component which encapsulates the filtering, something like:
```js
const filterFun = filter ? tg => tg.name?.toLowerCase().includes(filter) : () => true;
topicGroups.filter(tg => filterFun(tg) || tg.children.filter(filterFun)).map(tg => <TopicGroup ...);
```

src/discussions/topics/topic-group/TopicGroup.jsx
```diff
@@ -1,24 +1,68 @@
...
+      {name && (
+        <Link
+          className="topic-name list-group-item px-3 py-2 text-gray-300 text-decoration-none text-uppercase small"
+          to={
+            Routes.TOPICS.PATH.replace(':courseId', courseId)
+              .replace(':category?', name)
+          }
+        >
+          {name}
+        </Link>
```
kewne, line 53: This looks like a `TopicGroupLink` waiting be be extracted.

src/discussions/posts/post/Post.jsx
```diff
@@ -1,76 +1,164 @@
...
+
+function Post({
+  post,
+  intl,
+}) {
+  const dispatch = useDispatch();
+
+  return (
+    <div className="d-flex flex-column mt-2">
+      <PostHeader
```
kewne, line 106: Nit: Just passing in a single `post` prop makes this much smaller and doesn't really impact readability.

src/discussions/comments/data/slices.js
kewne: Can this be added as a comment or documentation?

I feel like encapsulating the map lookups would make this much easier to read and use.
For example, the `selectCommentResponses` and `selectThreadComments` selectors would become obsolete because you could instead use
```js
const selectComments = state => state.comments;

// then, in component...
const comments = useSelector(selectComments);
const responses = comments.getResponsesTo(commentId); // returns [] if there are no responses
```

src/discussions/comments/data/slices.js
```diff
@@ -3,14 +3,28 @@ import { createSlice } from '@reduxjs/toolkit';
...
-      }
-      if (!threads[comment.thread_id].includes(comment.id)) {
-        threads[comment.thread_id].push(comment.id);
+      if (comment.parent_id) {
+        if (!responses[comment.parent_id]) {
+          responses[comment.parent_id] = [];
+        }
+        if (!responses[comment.parent_id].includes(comment.id)) {
+          responses[comment.parent_id].push(comment.id);
+        }
```
kewne, line 19: Using sets makes this simpler IMO:
```js
if (!responses[comment.parent_id]) {
  responses[comment.parent_id] = new Set();
}
responses[comment.parent_id].add(comment.id);
```

src/discussions/comments/reply/Reply.jsx
```diff
@@ -0,0 +1,56 @@
...
+function Reply({
+  reply,
+  intl,
+}) {
+  const dispatch = useDispatch();
+  const hasParent = !!reply.parent_id;
+  const inlineReplies = useSelector(selectCommentResponses(reply.id));
+  useEffect(() => {
+    // If the comment has a parent comment, it won't have any children, so don't fetch them.
+    if (!hasParent) {
```
kewne, line 20: Seems more intuitive to name this `hasChildren`:
```js
const hasChildren = !reply.parent_id;
// ...
if (hasChildren) {
  dispatch(fetchCommentResponses(reply.id));
}
```

src/discussions/posts/PostsView.jsx
```diff
@@ -1,17 +1,80 @@
...
+  const loadingStatus = useSelector(threadsLoadingStatus());
+
+  let posts = [];
+  if (topicId) {
+    posts = useSelector(selectTopicThreads(topicId));
+  } else if (mine) {
+    posts = useSelector(selectUserThreads(authenticatedUser.username));
+  } else {
+    posts = useSelector(selectAllThreads());
+  }
```
kewne, line 41: This is a pretty weird construction but I couldn't find where the component is used; if it's easy to do, I'd pull this up to the parent.

src/discussions/topics/data/selectors.js
```diff
@@ -1,4 +1,33 @@
 /* eslint-disable import/prefer-default-export */
-export const selectCourseTopics = state => state.topics.topics;
+// eslint-disable-next-line no-unused-vars
+const filteredTopics = (topics, filter) => (filter
```
kewne, line 3: Is this used? It's not exported and is only used within itself in this file.

src/discussions/comments/reply/Reply.jsx
xitij2000: I think that's a lot cleaner. Will do.

kewne:
Woah, this was huge.
I'll :+1: this because the intent seems to be getting the basic structure down.

- [x] I tested this: mostly clicked around to check for broken things and went down the list of features
- [x] I read through the code

src/discussions/comments/data/slices.js
xitij2000: I don't think redux state can handle that. I do not think you can have arbitrary functions be part of redux state. Redux state should be plain javascript objects so that redux can reason about their state and deteming if/when the state has changed to propagate ui update events. 

Of course, we can encapsulate all the code to deal with the redux in a class, but I'm not sure how that helps. React/redux are generally more functional in paradigm than object-oriented and it'll be an ill fit.

src/discussions/comments/data/slices.js
xitij2000: I think I'm going to keep things the way they are. Making this a set means that we now can use `.map` on it like a regular array, so it will need to be converted to an array, or we'll need to use a different approach in the component. I'd rather that the components not need to be aware of the specific data structures inside the state.

src/discussions/posts/PostsView.jsx
xitij2000: This is the topmost view.

src/discussions/comments/data/slices.js
kewne: Maybe it can be embedded into the selector, then?

xitij2000: It is right? We ahev `selectCommentResponses` that takes a `commentId`, so you can use it as `useSelector(selectCommentResponses(commentId))`.

kewne: You're right!

awaisdar001 (commented):
Done the first pass basically I have a few questions here.

src/discussions/posts/index.js
```diff
@@ -1,2 +1,2 @@
 /* eslint-disable import/prefer-default-export */
-export { default as PostsViewContainer } from './PostsViewContainer';
+export { default as PostsView } from './PostsView';
```
mehaknasir, line 2: default imports

src/discussions/posts/post-actions-bar/messages.js
```diff
@@ -0,0 +1,16 @@
...
+    id: 'discussions.posts.action-bar.search',
+    defaultMessage: 'Search all posts',
+    description: 'Placeholder text in search box',
+  },
+  add_a_post: {
+    id: 'discussion.posts.action-bar.add',
+    defaultMessage: 'Add a post',
+    description: 'Button to add a new discussion post',
+  },
+});
```
mehaknasir, line 14: use camel case

src/discussions/posts/post-filter-bar/PostFilterBar.jsx
```diff
@@ -0,0 +1,102 @@
...
+            }
+          />
+        </div>
+        <div className="justify-content-end m-1">
+          <SelectableDropdown
+            selected={currentSorting}
+            options={postOrderingOptions}
+            onChange={(sortBy) => dispatch(setSortedBy(sortBy.value))}
+            label={
+              <FontAwesomeIcon icon={faSort} aria-label="Sort" title="Sort" />
```
mehaknasir, line 78: use paragon icons

public/index.html
```diff
@@ -1,10 +1,10 @@
...
       <meta charset="utf-8">
       <meta name="viewport" content="width=device-width, initial-scale=1.0">
-      <link rel="shortcut icon" href="/favicon.ico" type="image/x-icon" />
+      <link rel="shortcut icon" href="<%=htmlWebpackPlugin.options.FAVICON_URL%>" type="image/x-icon" />
```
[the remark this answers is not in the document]
xitij2000: I have updated this to match the current frotnend template: https://github.com/edx/frontend-template-application/blob/cacf080e87006473bb5da7d763c5eac2b7908750/public/index.html#L7

src/discussions/posts/post-filter-bar/messages.js
```diff
@@ -0,0 +1,81 @@
...
+    defaultMessage: 'All questions',
+    description: 'Option in dropdown to filter to all questions',
+  },
+  filter_by: {
+    id: 'discussions.posts.filter.message',
+    defaultMessage: 'Status: {filterBy}',
+    description: 'Display text used to indicate what post status is being filtered',
+  },
+  filter_all: {
+    id: 'discussions.posts.status.filter.all',
```
mehaknasir, line 25: use camel case

src/discussions/posts/post/Post.jsx
```diff
@@ -87,6 +37,120 @@ export const postShape = PropTypes.shape({
   updated_at: PropTypes.string,
```
mehaknasir, line 37: use normalize data here

src/discussions/comments/reply/Reply.jsx
```diff
@@ -0,0 +1,56 @@
...
+            <span className="font-weight-bold text-info-300 mr-1">
+              {reply.author}
+            </span>
+            <span title={reply.created_at}>
+              {intl.formatMessage(messages.reply_time, {
+                relativeTime: timeago.format(reply.created_at, intl.locale),
+              })}
+            </span>
+          </div>
+          <div className="reply-body d-flex" dangerouslySetInnerHTML={{ __html: reply.rendered_body }} />
```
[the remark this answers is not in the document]
xitij2000: This is the only way you can take existing HTML content from an API and inject it in an React view. 
Ref: https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml

src/discussions/posts/post/Post.jsx
```diff
@@ -87,6 +37,120 @@ export const postShape = PropTypes.shape({
...
 });
 
+function PostTypeIcon(props) {
+  return (
+    <div className="m-1">
+      {props.type === 'question' && <FontAwesomeIcon icon={faQuestionCircle} size="lg" />}
+      {props.type === 'discussion' && <FontAwesomeIcon icon={faComments} size="lg" />}
+      {props.pinned && (
+        <FontAwesomeIcon
```
mehaknasir, line 46: use from paragon

src/discussions/posts/post/Post.jsx
```diff
@@ -87,6 +37,120 @@ export const postShape = PropTypes.shape({
...
+          </span>
+          <span className="d-flex small text-gray-300">
+            {post.author}{' | '}
+            <span title={post.created_at}>
+              {intl.formatMessage(messages.posted_on, { time: timeago.format(post.created_at, intl.locale) })}
+            </span>
+          </span>
+        </div>
+        <div className="d-flex align-items-center mr-3">
+          <FontAwesomeIcon icon={faComments} className="mr-2" />
```
mehaknasir, line 87: use icon from paragon

src/discussions/posts/post/PostLink.jsx
```diff
@@ -0,0 +1,42 @@
...
+      className="discussion-post list-group-item list-group-item-action p-0 text-decoration-none text-gray-900"
+      to={
+        Routes.COMMENTS.PATH.replace(':courseId', post.course_id)
+          .replace(':topicId', post.topic_id)
+          .replace(':postId', post.id)
+      }
+    >
+      {post.abuse_flagged && (
+        <div className="bg-gray-100 flex-fill">
+          <FontAwesomeIcon icon={faFlag} className="mx-1" />
```
mehaknasir, line 25: avoid font awesome and use paragon icons

src/discussions/posts/post/PostLink.jsx
```diff
@@ -0,0 +1,42 @@
...
+          {intl.formatMessage(messages.content_reported)}
+        </div>
+      )}
+      <div className="d-flex flex-row mb-2">
+        <FontAwesomeIcon icon={faCircle} className={`my-1 text-accent-a ${post.read && 'invisible'}`} size="xs" />
```
mehaknasir, line 30: avoid font awesome and use paragon icons

src/discussions/posts/post/messages.js
```diff
@@ -9,6 +9,30 @@ const messages = defineMessages({
...
+  like: {
+    id: 'discussions.post.like',
+    defaultMessage: 'Like',
+    description: 'Tooltip/alttext for button to like a discussion post',
+  },
+  remove_like: {
+    id: 'discussions.post.remove-like',
+    defaultMessage: 'Remove like',
+    description: 'Tooltip/alttext for button to remove the like applied to a discussion post',
+  },
```
mehaknasir, line 35: use camel case

src/discussions/topics/data/selectors.js
```diff
@@ -1,4 +1,33 @@
...
+export const selectCourseTopic = topicId => state => state.topics.topics.courseware_topics[topicId]
+  || state.topics.topics.non_courseware_topics[topicId];
+
+export const selectTopicCategory = topicId => state => (
+  state.topics.topics.courseware_topics.find(category => (
+    category.children.find(topic => topic.id === topicId)
+  ))
+);
+
+export const topicsLoadingStatus = state => (
```
mehaknasir, line 31: you can use like this to remove lint warning
```
 default export {
  selectTopicCategory,
  topicsLoadingStatus,
........ etc
}
```

src/discussions/posts/data/slices.js
```diff
@@ -32,13 +48,20 @@ const threadsSlice = createSlice({
...
   },
   reducers: {
     fetchThreadsRequest: (state) => {
       state.status = RequestStatus.IN_PROGRESS;
     },
     fetchThreadsSuccess: (state, { payload }) => {
       state.status = RequestStatus.SUCCESSFUL;
+      state.topicThreadMap = {};
+      state.threads = {};
       normaliseThreads(state, payload.results);
```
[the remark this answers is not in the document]
xitij2000: >  I would normalize before putting data into the reducer. It **should not** be done in the reducer, imho

Could you elaborate?

We're building up the state based on API responses from multiple different APIs, and they are all inter-related, and the API code doesn't really know the broader state it needs to act on. So, the reducer seemed like the most natural place since it has all the data. 

I can definitely move it if that's preferable. 

>  Also, the reducer is manipulating state directly, which is a redux anti-pattern. It should create a New state object. Manipulating state in place like that can cause all sorts of issues.

The state is **not** being manipulated in place. I've been using React and Redux long enough to know not to do that :-)
What's happening here is that this project, and other Open edX frontends use a library called [redux toolkit](https://redux-toolkit.js.org/), which is where the [createSlice](https://redux-toolkit.js.org/api/createSlice) function is coming from. This library automatically creates a new state object for you so you can manipulate with comfort and the state remains immutable.

You can also see it in action [here for instance](https://github.com/edx/frontend-app-learning/blob/master/src/course-home/data/slice.js)

src/discussions/topics/data/slices.js
```diff
@@ -1,15 +1,17 @@
 /* eslint-disable no-param-reassign,import/prefer-default-export */
```
mehaknasir, line 1: if you will export default in constants file you wont have to suppress prefer default export here

src/discussions/topics/data/slices.js
```diff
@@ -1,15 +1,17 @@
...
+import { RequestStatus, TopicOrdering } from '../../../data/constants';
 
-const topicsSLice = createSlice({
+const topicsSlice = createSlice({
   name: 'courseTopics',
   initialState: {
     status: RequestStatus.IN_PROGRESS,
     topics: {
       courseware_topics: [],
       non_courseware_topics: [],
```
mehaknasir, line 11: camel case

src/discussions/topics/index.js
```diff
@@ -1,2 +1,2 @@
 /* eslint-disable import/prefer-default-export */
-export { default as TopicsViewContainer } from './TopicsViewContainer';
+export { default as TopicsView } from './TopicsView';
```
mehaknasir, line 2: use default exports

src/discussions/topics/topic-group/topic/Topic.jsx
```diff
@@ -1,58 +1,69 @@
...
+          {questions}
+        </div>
+        <div className="badge mr-4">
+          <FontAwesomeIcon className="mr-2" icon={faComments} />
+          {discussions}
+        </div>
+        {flags!==null && (
+          <div className="badge">
+            <FontAwesomeIcon className="mr-2" icon={faFlag} />
+            {flags}
```
mehaknasir, line 44: use paragon icons

src/discussions/topics/topic-search-bar/TopicSearchBar.jsx
```diff
@@ -0,0 +1,39 @@
...
+        className="flex-fill m-1 border-0"
+        placeholder={intl.formatMessage(messages.find_a_topic)}
+        onSubmit={(query) => dispatch(setFilter(query))}
+        onChange={(query) => dispatch(setFilter(query))}
+      />
+      <SelectableDropdown
+        options={topicOrderingOptions}
+        onChange={(option) => dispatch(setSortBy(option.value))}
+        label={
+          <FontAwesomeIcon icon={faSort} />
```
mehaknasir, line 28: use icon from paragon

src/discussions/topics/topic-search-bar/TopicSearchBar.jsx
```diff
@@ -0,0 +1,39 @@
...
+    <div className="d-flex flex-row p-1 align-items-center">
+      <SearchField
+        className="flex-fill m-1 border-0"
+        placeholder={intl.formatMessage(messages.find_a_topic)}
+        onSubmit={(query) => dispatch(setFilter(query))}
+        onChange={(query) => dispatch(setFilter(query))}
+      />
+      <SelectableDropdown
+        options={topicOrderingOptions}
+        onChange={(option) => dispatch(setSortBy(option.value))}
```
mehaknasir, line 26: seems like it will send an API call on every key change. please use debouncing rule here

src/discussions/topics/topic-search-bar/messages.js
```diff
@@ -0,0 +1,31 @@
...
+    id: 'discussions.topics.sort.comment-count',
+    defaultMessage: 'Most activity',
+    description: 'Option in dropdown to sort topics by most activity',
+  },
+  sort_by_course_structure: {
+    id: 'discussions.topics.sort.course-structure',
+    defaultMessage: 'Course Structure',
+    description: 'Option in dropdown to sort topics by course structure',
+  },
+  find_a_topic: {
```
mehaknasir, line 24: use camel case

src/index.scss
mehaknasir: I think we dont need font-awesome

mehaknasir (changes requested):
as a general note have already seen snake case at many places can we please make sure we are following JS convention
import sequence is not correct in many places. use default imports instead of named imports.
Use paragon icons instead of font awesome

src/components/SelectableDropdown.jsx
```diff
@@ -42,7 +42,6 @@ SelectableDropdown.propTypes = {
       label: PropTypes.string,
     }),
   ).isRequired,
-  defaultOption: PropTypes.string.isRequired,
   onChange: PropTypes.func,
   label: PropTypes.node,
```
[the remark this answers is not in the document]
xitij2000: I'm not sure what you mean, the default props are defined in the very next block of code.

src/discussions/comments/comment-icons/CommentIcons.jsx
```diff
@@ -0,0 +1,46 @@
...
+import Button from '@edx/paragon/dist/Button';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+function CommentIcons({ abuseFlagged, following }) {
+  return (
+    <div className="d-flex flex-column icons">
+      {/* Only show the star if the comment has a following attribute, indicating it can be followed */}
+      { following !== undefined && (
+        following
```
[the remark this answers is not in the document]
xitij2000: I think you're misunderstanding the logic here. If following is undefined, i.e. the attribute doesn't exist, then we don't show the star at all. If the attribute does exist, then we show an empty or hollow star depending on its value.

src/discussions/comments/comment/Comment.jsx
```diff
@@ -1,56 +1,51 @@
...
-      </div>
-      <div className="comment-body d-flex" dangerouslySetInnerHTML={{ __html: comment.rendered_body }} />
-      <div className="visibility-comment d-flex">
-        {/*  TODO: Add parent group info */ }
+        <CommentIcons abuseFlagged={comment.abuse_flagged} following={comment.following} />
       </div>
-      <div className="actions d-flex">
-        <Button>{ intl.formatMessage(messages.add_response) }</Button>
+      <div className="mt-2">
+        <div className="d-flex" dangerouslySetInnerHTML={{ __html: comment.rendered_body }} />
```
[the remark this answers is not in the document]
xitij2000: That library is no longer maintained. 
That said, this code is a temporary way to get this working. What we should probably do is to use the markdown version and convert that into react using something like https://www.npmjs.com/package/react-markdown

That will also be useful later when we have to show users an HTML preview.

src/components/index.js
```diff
@@ -1 +1,3 @@
-export * from './selectable-dropdown';
+/* eslint-disable import/prefer-default-export */
+export { default as SelectableDropdown } from './SelectableDropdown';
+export { default as PostActionsBar } from '../discussions/posts/post-actions-bar/PostActionsBar';
```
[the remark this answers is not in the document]
xitij2000: This will not work, and I don't think it's valid export syntax. You can read more here: https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export#re-exporting_aggregating

src/discussions/comments/comment-icons/CommentIcons.jsx
```diff
@@ -0,0 +1,46 @@
...
+          )
+      )}
+      { abuseFlagged && (
+        <Button variant="link" className="p-0" size="xs">
+          <FontAwesomeIcon icon={faFlag} />
+        </Button>
+      ) }
+      <Button variant="link" className="p-0" size="xs">
+        <FontAwesomeIcon icon={faEllipsisV} />
+      </Button>
```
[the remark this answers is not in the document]
xitij2000: Not all the icons we need are available on paragon yet. Instead of partially replacing them and ending up with an inconsistent UI and potential issues due to different sizing, perhaps we can leave this for a future PR when all the icons are available? 
Does that sound OK?

src/discussions/comments/messages.js
```diff
@@ -0,0 +1,32 @@
...
+    defaultMessage: `{postType, select, 
+      discussion {Discussion} 
+      question {Question}
+    } posted {relativeTime} by`,
+    description: 'Timestamp for when a user posted the message followed by username. The relative time is already translated.',
+  },
+  reply_time: {
+    id: 'discussions.comments.comment.replied_time',
+    defaultMessage: 'posted {relativeTime}',
+    description: 'Message about hwo long ago a reply was posted. Appears as "username posted 7 minutes ago"',
```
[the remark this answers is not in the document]
xitij2000: According to the current i18n docs [here](https://github.com/edx/frontend-platform/blob/master/docs/how_tos/i18n.rst) this should be the same as the ID. This is also used by a number of other apps. 
Which convention should I follow?

src/discussions/navigation/breadcrumb-menu/BreadcrumbMenu.jsx
```diff
@@ -0,0 +1,58 @@
...
+  return (
+    <div className="breadcrumb-menu d-flex flex-row mt-2 mx-3">
+      <Breadcrumb
+        links={crumbs}
+        activeLabel={activeLabel}
+        spacer={<span className="custom-spacer">/</span>}
+        clickHandler={(e) => {
+          e.preventDefault();
+          history.push(e.target.pathname);
+          return true;
```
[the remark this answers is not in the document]
xitij2000: Now that we have preventDefault, I don't think it's needed, will remove.

src/discussions/comments/comment-icons/CommentIcons.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+CommentIcons.propTypes = {
+  abuseFlagged: PropTypes.bool,
+  following: PropTypes.bool,
+};
+
+CommentIcons.defaultProps = {
+  abuseFlagged: undefined,
+  following: undefined,
+};
```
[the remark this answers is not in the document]
xitij2000: An undefined value for a bool can indicate that that data isn't available, whereas a value of false indicates that the data is available. 

In this case, if we can use the component is a number of ways:
- following =  true, show a full star
- following = false, show an empty star
- following = undefined, don't show a star.

src/discussions/topics/topic-search-bar/TopicSearchBar.jsx
mehaknasir: okay I think its only updating the store and no API call is made so we are good here with this.

src/discussions/topics/topic-group/topic/Topic.jsx
awaisdar001: I think we do not (yet) have these icons in Paragon, so what we can do is to have a ticket in the backlog to replace those with paragon icons and remove the dependency of FA. This shouldn't be a blocker.

src/components/index.js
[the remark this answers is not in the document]
mehaknasir: Thank you for raising this point. I will resolve all these reviews.

src/discussions/posts/data/slices.js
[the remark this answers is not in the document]
awaisdar001: Thanks for the context regarding redux toolkit.

Please do not normalize the data in the reducer, the reducer is only suppose to update state not change / update any given state. If you can update, that would be great. thank you.

src/discussions/comments/comment-icons/CommentIcons.jsx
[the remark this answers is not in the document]
mehaknasir: Okay, got your logic now.

src/discussions/comments/comment-icons/CommentIcons.jsx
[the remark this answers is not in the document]
mehaknasir: sounds good to me.

src/discussions/comments/comment-icons/CommentIcons.jsx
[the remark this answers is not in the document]
mehaknasir: okay. got the logic now.

src/discussions/comments/comment/Comment.jsx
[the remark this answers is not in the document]
mehaknasir: Okay, makes sense to me.

src/discussions/comments/messages.js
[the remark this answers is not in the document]
mehaknasir: when you will normalize data before saving it in the store the keys will automatically be in camel case and hence you will need to change all these keys in the messages file to camel case I think.

src/discussions/topics/topic-search-bar/TopicSearchBar.jsx
xitij2000: In either case, this is a dropdown, so can't really debounce, if there is a change, it need to be reflected.

package.json
```diff
@@ -34,34 +35,36 @@
...
-    "@edx/frontend-platform": "^1.7.0",
-    "@edx/paragon": "^12.0.5",
-    "@fortawesome/fontawesome-svg-core": "^1.2.32",
-    "@fortawesome/free-brands-svg-icons": "^5.15.1",
-    "@fortawesome/free-regular-svg-icons": "^5.15.1",
-    "@fortawesome/free-solid-svg-icons": "^5.15.1",
-    "@fortawesome/react-fontawesome": "^0.1.12",
-    "@reduxjs/toolkit": "^1.4.0",
+    "@edx/brand": "npm:@edx/brand-openedx@1.1.0",
+    "@edx/frontend-component-footer": "^10.1.4",
```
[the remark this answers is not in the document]
xitij2000: I've pinned to exact versions now.

src/discussions/posts/data/slices.js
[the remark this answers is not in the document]
xitij2000: Sure, will do. It's a biggish change so won't be able to get to it today.

src/components/SelectableDropdown.jsx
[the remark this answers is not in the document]
awaisdar001: its ok if you want to have a separate PR for that. thoughts?

src/discussions/posts/data/api.js
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable import/prefer-default-export */
-import { ensureConfig, getConfig } from '@edx/frontend-platform';
+import snakeCase from 'lodash.snakecase';
+import { ensureConfig, getConfig, snakeCaseObject } from '@edx/frontend-platform';
```
mehaknasir, line 3: May I know the reason when we have `snakeCaseObject ` then why we are adding lodash package again and using both of these in same file?

xitij2000: The reason for this is that snakeCaseObject only converts the **keys** of an object from snake case to camel case, however in this case we also need to modify some values.

mehaknasir: I see, I think we should have that utility available from frontend-platform like other utilities, but I searched there is none.
I can discuss with this team further for now it makes sense.

xitij2000: If it helps, `frontend-platform` uses the same function: https://github.com/edx/frontend-platform/blob/1e062a3579842c5adcbf8f8f5953a1f51f24e994/src/utils.js#L6-L7

That's where I looked, so I copied over the dependency. I think what we should do is to just reexport that from frontend-platform so it's available to import from there. That way a future version of the platform can replace lodash with a different implementation.

awaisdar001: I have discussed this within the team and they think you should be fine using "loadash" as a dependency here.  frontend-platform doesn’t bundle its own dependencies, so you should only get one copy of the lodash functions.

mehaknasir: marking it resolved no change needed.

src/discussions/topics/topic-group/topic/Topic.jsx
mehaknasir: I think Icons are now available but we decided to refactor icons in a separate PR. hence marking it resolve

src/discussions/comments/messages.js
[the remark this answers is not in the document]
mehaknasir: We are in process of updating those docs to avoid any future misunderstandings.

src/discussions/comments/data/api.js
```diff
@@ -15,43 +15,41 @@ const commentsApiUrl = `${apiBaseUrl}/api/discussion/v1/comments/`;
...
 ) {
   const url = `${commentsApiUrl}${commentId}/`;
   const params = {
     page,
     page_size: pageSize,
-    requested_fields: requestedFields,
+    requested_fields: 'profile_image',
   };
   const { data } = await getAuthenticatedHttpClient().get(url, { params });
   return data;
```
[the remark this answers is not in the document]
awaisdar001: `camelCaseObject` normalize data here.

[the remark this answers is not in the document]
xitij2000: I'm normalising data in the thunk instead, so this code returns the raw results.

src/discussions/navigation/breadcrumb-menu/BreadcrumbMenu.jsx
```diff
@@ -0,0 +1,58 @@
...
+  const history = useHistory();
+
+  const crumbs = [
+    {
+      url: Routes.TOPICS.ALL.replace(':courseId', courseId),
+      label: intl.formatMessage(messages.all_topics),
+    },
+    {
+      url: Routes.TOPICS.CATEGORY.replace(':courseId', courseId)
+        .replace(':category', category || (topicCategory && topicCategory.name)),
```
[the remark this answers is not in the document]
xitij2000: The entire breadcrumb menu will be rewritten since the new design is very different, so I'm avoiding changes here.

src/discussions/posts/data/selectors.js
```diff
@@ -1,6 +1,21 @@
...
+
+export const selectUserThreads = author => state => (
+  Object.values(state.threads.threads)
+    .filter(thread => thread.author === author)
+);
+
+export const selectThreadSorting = () => state => state.threads.sortedBy;
+export const selectThreadStatusFilter = () => state => state.threads.statusFilter;
+export const selectThreadAllPostsTypeFilter = () => state => state.threads.allTypeFilter;
+export const selectThreadMyPostsTypeFilter = () => state => state.threads.myTypeFilter;
```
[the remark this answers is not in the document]
xitij2000: I think that kind of thing is tying the React component to the state model making it harder to change that model in the future.

src/discussions/posts/post-filter-bar/PostFilterBar.jsx
```diff
@@ -29,9 +24,11 @@ function PostFilterBar({
 }) {
   const dispatch = useDispatch();
   const currentSorting = useSelector(selectThreadSorting());
-  const currentStatusFilter = useSelector(selectThreadStatusFilter());
-  const currentAllPostsTypeFilter = useSelector(selectThreadAllPostsTypeFilter());
-  const currentMyPostsTypeFilter = useSelector(selectThreadMyPostsTypeFilter());
+  const {
+    statusFilter: currentStatusFilter,
+    allTypeFilter: currentAllPostsTypeFilter,
+    myTypeFilter: currentMyPostsTypeFilter,
```
mehaknasir, line 30: I think if we give them meaningful name in reducer we wont be needing it to rename here

src/discussions/navigation/breadcrumb-menu/BreadcrumbMenu.jsx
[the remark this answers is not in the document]
mehaknasir: do we have any ticker for that. just want to clear my self.

src/discussions/posts/data/slices.js
```diff
@@ -1,18 +1,35 @@
...
-      if (!topics[thread.topic_id].includes(thread.id)) {
-        topics[thread.topic_id].push(thread.id);
+      if (!topics[thread.topicId].includes(thread.id)) {
+        topics[thread.topicId].push(thread.id);
       }
-      threads[thread.id] = thread;
+      threads[thread.id] = normaliseProfileImage(threads[thread.id], thread);
     },
   );
 }
```
mehaknasir, line 35: marking

mehaknasir, line 35: I think this will be changed with normalization changes. making it as a todo item

src/discussions/navigation/breadcrumb-menu/BreadcrumbMenu.jsx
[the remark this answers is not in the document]
mehaknasir: marking it to do itema created a ticket for that.

src/discussions/posts/post-filter-bar/PostFilterBar.jsx
xitij2000: I don't think that's true here. If you're dealing with two filter values, then you do need to disambiguate them somehow. Here I'm adding current to clarify that it's the current value. In most other places that implicit.

mehaknasir (approved):
I have reviewed the PR and approving this but it has some to-do items which are added as tickets to Jira already and are assigned to @xitij2000. Have a look.

https://openedx.atlassian.net/browse/TNL-8528
https://openedx.atlassian.net/browse/TNL-8519
https://openedx.atlassian.net/browse/TNL-8517

PR #12, opened 2021-08-31
tests: Clean up and optimise redux data layer and add tests [BD-38] [TNL-8517] [BB-4643]

xitij2000:
This change adds tests for the redux data store, moves the normalization of data out of the reducer, and reorganizes data for better access.

**Tickets:** [TNL-8517](https://openedx.atlassian.net/browse/TNL-8517)
**Dependencies**: None

**Screenshots**: NA

**Merge deadline**: "None"

**Testing instructions**:

1. Run the test suite
2. Browse around the app and make sure everything works as expected. 

**Reviewers**
- [ ] (OpenCraft internal reviewer's GitHub username goes here)
- [ ] edX reviewer[s] TBD


edx-webhook:
Thanks for the pull request, @xitij2000! I've created [BLENDED-948](https://openedx.atlassian.net/browse/BLENDED-948) to keep track of it in Jira.   More details are on the [BD-38](https://openedx.atlassian.net/wiki/spaces/COMM/pages/2790490201/BD-38+Discussions+In-Context+Discussion+Moderation+Phase+2) project page.


.eslintrc.js
```diff
@@ -1,3 +1,10 @@
 const { createConfig } = require('@edx/frontend-build');
 
-module.exports = createConfig('eslint');
+module.exports = createConfig('eslint',
+  {
+    'rules': {
+      'import/no-extraneous-dependencies': 'off',
```
xitij2000, line 6: I've added this since otherwise `rosie`, which is a devDependency is incorrectly recommended as a regular dependency since it's imported in a non-test file. This can be fixed in `edx-lint`.

tinumide:
:+1:

- [x] I tested this: Ran the test suite and browsed through the app to make sure everything works as expected
- [x] I read through the code
- [x] ~Includes documentation~ NA

tinumide (approved):
:+1:

- [x] I tested this: Ran the test suite and browsed through the app to make sure everything works as expected
- [x] I read through the code
- [ ] ~Includes documentation~ NA

src/discussions/posts/data/thunks.js
```diff
@@ -67,7 +106,8 @@ export function fetchThreads(courseId, {
     try {
       dispatch(fetchThreadsRequest({ courseId }));
       const data = await getThreads(courseId, options);
-      dispatch(fetchThreadsSuccess(camelCaseObject(data)));
+      const normalised = normaliseThreads(camelCaseObject(data));
```
mehaknasir, line 109: just a suggestion to rename this variable named **normalised** to **normalisedData**? maybe or if you have any more accurate name in mind?

src/discussions/comments/data/slices.js
```diff
@@ -59,8 +30,10 @@ const commentsSlice = createSlice({
     },
     fetchCommentsSuccess: (state, { payload }) => {
       state.status = RequestStatus.SUCCESSFUL;
-      normaliseComments(state, payload.results);
-      state.page = payload.pagination.page;
+      state.pages[payload.page] = payload.ids;
+      Object.assign(state.commentsInThreads, payload.commentsInThreads);
+      Object.assign(state.commentsInComments, payload.commentsInComments);
+      Object.assign(state.commentsById, payload.commentsById);
```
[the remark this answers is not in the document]
xitij2000: I find this a bit cleaner than the spread operator in this case, which would be roughly:

```state.commentsById = {...state.commentsById, ...payload.commentsById};```

File: src/discussions/posts/data/thunks.js (hunk 2 above)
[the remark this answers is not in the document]
Comment (xitij2000) [reply src/discussions/posts/data/thunks.js:109]: Sure, I think I just added this variable to print the data while testing things. I can just eliminate it and pass it in directly.

File: .eslintrc.js [hunk 4]
```diff
@@ -1,3 +1,28 @@
...
+            // Non-react third-party packages come next
+            ['^@?\\w'],
+            // Packages from the @edx namespace come after that
+            ['^@edx?\\w'],
+            // Finally we have internal, relative imports
+            ['^\\.\\.(?!/?$)', '^\\.\\./?$', '^\\./(?=.*/)(?!/?$)', '^\\.(?!/?$)', '^\\./?$'],
+          ],
+        },
+      ],
+      'simple-import-sort/exports': 'error',
```
xitij2000, line 24: This plugin will make sure that the imports are sorted.

src/discussions/comments/data/selectors.js
```diff
@@ -1,10 +1,10 @@
 /* eslint-disable import/prefer-default-export */
-export const selectThreadComments = threadId => state => (state.comments.threadCommentMap[threadId] || []).map(
-  commentId => state.comments.comments[commentId],
+export const selectThreadComments = threadId => state => (state.comments.commentsInThreads[threadId] || []).map(
```
awaisdar001, line 2: Review-Reminder: Test `selectThreadComments(1) === selectThreadComments(1)`

src/discussions/posts/data/selectors.js
```diff
@@ -1,20 +1,36 @@
 /* eslint-disable import/prefer-default-export */
 
-export const selectTopicThreads = topicId => state => (state.threads.topicThreadMap[topicId] || []).map(
-  threadId => state.threads.threads[threadId],
+export const selectTopicThreads = topicId => state => (state.threads.threadsInTopic[topicId] || []).map(
+  threadId => state.threads.threadsById[threadId],
 );
 
-export const selectThread = threadId => state => (state.threads.threads?.[threadId]);
+export const selectThread = threadId => state => (state.threads.threadsById?.[threadId]);
```
awaisdar001, line 7: You are not using `useModel` / `addModel` pattern here - whichi I guess most of the MFE use including course-authoring. Can you explain why?

xitij2000: The hierarchy is more complex here, and isn't repeated across multiple models. So the existing useModel framework doesn't work well here. 

Currently, useModel will just save objects by id, however it doesn't care about other id references, for example, threadId or parentId.  Here we need to look up comments by their id, but also by their thread id or parent id. 

I've used [normalizr](https://github.com/paularmstrong/normalizr) for things like this before, and that might be another way to go.

mehaknasir: We looked at this normalizr implementation together and I think we can go with this and see how clean code gets after using it.

mehaknasir:
Also, do we have a ticket for this task where we can track the progress? If yes kindly add it in descriptipn.

xitij2000:
> Also, do we have a ticket for this task where we can track the progress? If yes kindly add it in descriptipn.

The ticket number is in the PR title.

xitij2000:
@awaisdar001 While I did mention normalizr as an option, I think it's a bit too late to use that. 

Switching to normalizr at this point is a pretty big change, and I don't see any benefit since it'll just put the data in the same shape as it currently is. It's also not used by any of the other frontends. I'm reluctant to perform a major refactor at the end of a task that was already a major refactor for very little benefit. 

Switching will also mean further changes to how we handle pagination etc.

mehaknasir (commented):
> @awaisdar001 While I did mention normalizr as an option, I think it's a bit too late to use that.
> 
> Switching to normalizr at this point is a pretty big change, and I don't see any benefit since it'll just put the data in the same shape as it currently is. It's also not used by any of the other frontends. I'm reluctant to perform a major refactor at the end of a task that was already a major refactor for very little benefit.
> 
> Switching will also mean further changes to how we handle pagination etc.

Hi @xitij2000,
it's alright if you don't want to use normalizr at this stage. but there are a few points to notice. normalizr will not change redux structure and will only help us clean the code. secondly, pagination is not implemented right now so I don't get how it will impact pagination.

src/discussions/comments/data/slices.js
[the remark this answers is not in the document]
mehaknasir: we discussed it internally and we feel like we should follow the redux suggested syntax which follows spread operator syntax

src/discussions/posts/data/thunks.js
```diff
@@ -67,7 +106,8 @@ export function fetchThreads(courseId, {
     try {
       dispatch(fetchThreadsRequest({ courseId }));
       const data = await getThreads(courseId, options);
-      dispatch(fetchThreadsSuccess(camelCaseObject(data)));
+      const normalisedData = normaliseThreads(camelCaseObject(data));
```
mehaknasir, line 109: Everywhere in the code base, we are normalizing data in API actions. what do you think is the best place to normalize data in thunks or in API actions?

xitij2000: I think thunks are a better place since it's possible to have two separate thunks that use the same kind of API differently and need different kinds of normalisation. So I try to keep the API side very raw and make it just a mapping of an API to function. The thunks then convert that into an operation on state. So I can imagine a thunk doing multiple API calls for instance, if needed and normalising them together.

src/discussions/comments/data/selectors.js
```diff
@@ -1,10 +1,10 @@
...
+  commentId => state.comments.commentsById[commentId],
 );
 
-export const selectCommentResponses = commentId => state => (state.comments.commentResponsesMap[commentId] || []).map(
-  cId => state.comments.comments[cId],
+export const selectCommentResponses = commentId => state => (state.comments.commentsInComments[commentId] || []).map(
+  cId => state.comments.commentsById[cId],
 );
```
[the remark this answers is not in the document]
xitij2000: I am a fan of reselet so no issues using it. I just avoid introducing new libraries that aren't already used. If this is OK, I can definitely clean up the seletors with it.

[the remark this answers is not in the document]
mehaknasir: I think reselect is available in the redux toolkit so you won't have to add another library for that.

src/discussions/posts/data/thunks.js
mehaknasir: Agreed. 👍

src/discussions/comments/data/selectors.js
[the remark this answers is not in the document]
xitij2000: I've updated the code to use reselect. That said we currently can't take advantage of most of its feature related to memorization since we'll be creating selectors dynamically.

src/discussions/comments/data/selectors.js
```diff
@@ -1,10 +1,23 @@
...
+    selectCommentsById,
+  ],
+  mapIdToComment,
 );
 
-export const selectCommentResponses = commentId => state => (state.comments.commentResponsesMap[commentId] || []).map(
-  cId => state.comments.comments[cId],
+export const selectCommentResponses = commentId => createSelector(
+  [
+    state => state.comments.commentsInComments[commentId] || [],
```
mehaknasir, line 17: why do you need this empty array here I think the way normalization is handled in thunk this will never be undefined.

src/discussions/comments/data/slices.js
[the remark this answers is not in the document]
mehaknasir: any update on this? so we can finalize this PR.

mehaknasir (commented):
there are only a few points left on this PR after that will be ready for merge can you please take a look. Thanks

src/discussions/comments/data/slices.js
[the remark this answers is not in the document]
xitij2000: @mehaknasir Sure. I can update this, but FYI, if you're concerned about the redux recommendations, they themselves use `Object.assign` in their section on mutating state [here](https://redux-toolkit.js.org/usage/immer-reducers#mutating-and-returning-state) as an alternative to using the spread operator, which makes the code verbose.

src/discussions/comments/data/selectors.js
xitij2000: If you refresh the page on a comment that hasn't loaded yet, this array will be empty. So instead of breaking the page, or needing to add logic for handling nulls everywhere, we return an empty array in that case.

xitij2000:
@mehaknasir  This is ready for another review pass.

src/discussions/posts/data/slices.js
```diff
@@ -65,10 +43,10 @@ const threadsSlice = createSlice({
     },
     fetchThreadsSuccess: (state, { payload }) => {
       state.status = RequestStatus.SUCCESSFUL;
-      state.topicThreadMap = {};
-      state.threads = {};
-      normaliseThreads(state, payload.results);
-      state.page = payload.pagination.page;
+      state.pages[payload.page - 1] = payload.ids;
```
mehaknasir, line 46: Pages are initialized as an empty array but here it is being changed to object with keys starting from zero. I am unable to comprehend this logic and I see no use of this anywhere. Are you sure this change is part of this PR?

src/discussions/posts/data/slices.js
```diff
@@ -97,7 +76,11 @@ const threadsSlice = createSlice({
     },
     postThreadSuccess: (state, { payload }) => {
       state.postStatus = RequestStatus.SUCCESSFUL;
-      normaliseThreads(state, [payload]);
+      state.threadsById[payload.id] = payload;
+      state.threadsInTopic[payload.topicId].push(payload.id);
+      // Temporarily add it to the top of the list so it's visible
+      state.pages[1] = [payload.id] + (state.pages[1] || []);
```
mehaknasir, line 82: this is now not making sense with the above logic.

src/discussions/posts/data/slices.js
```diff
@@ -127,10 +111,11 @@ const threadsSlice = createSlice({
     },
     deleteThreadSuccess: (state, { payload }) => {
       const { threadId } = payload;
-      const { topicId } = state.threads[threadId];
+      const { topicId } = state.threadsById[threadId];
       state.postStatus = RequestStatus.SUCCESSFUL;
-      state.topicThreadMap[topicId] = state.topicThreadMap[topicId].filter(item => item !== threadId);
-      delete state.threads[threadId];
+      state.threadsInTopic[topicId] = state.threadsInTopic[topicId].filter(item => item !== threadId);
+      state.pages = state.pages.map(page => page?.filter(item => item !== threadId));
```
mehaknasir, line 117: this will also not work as pages are being altered to object while fetching.

mehaknasir (changes requested):
Pages logic is not comprehensible. Its implication is not visible in PR as well. kindly suggest how can this be resolved

src/discussions/posts/data/slices.js
xitij2000: @mehaknasir It is not being changed to an object with numeric keys starting from zero. It is still an array. 
You can see this used in the selectors when ids in pages are mapped to items in them.

src/discussions/posts/data/slices.js
xitij2000: I'm not sure I understand why you think pages is being altered to an object? It is still an array, and the tests dealing with this code are still passing.

src/discussions/posts/data/slices.js
xitij2000: Fixed.

xitij2000:
> Pages logic is not comprehensible. Its implication is not visible in PR as well. kindly suggest how can this be resolved

I'm not sure I understand why it's not comprehensible? 
Earlier I was using an object which was indexed by page numbers as keys. I've now changed that to use an array instead, because an array is also something you can index using numbers, but provides features like map, flatMap etc.

awaisdar001 (approved):

src/discussions/posts/data/slices.js
mehaknasir: I confirm it again. now its being converted to array. resolving this.

mehaknasir (approved):

src/discussions/posts/data/slices.js
mehaknasir: Thanks. it resolves my concern

edx-webhook:
@xitij2000 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/frontend-app-discussions&entry.867055334=https://github.com/edx/frontend-app-discussions/pull/12&entry.1484655318=https://github.com/xitij2000&entry.752974735=2021-08-31+12:33&entry.1917517419=2021-09-15+10:40&entry.2133058324=Yes) so we can improve your experience in the future.

PR #21, opened 2021-09-16
Add post banners [BD-38] [BB-4737] [TNL-8655]

tinumide:
## Description

This PR Implement post banners for:

- Pinned posts
- Endorsed posts
- Reported posts
- closed posts

## Supporting Information

[BB-4737](https://tasks.opencraft.com/browse/BB-4737)
[TNL-8655](https://openedx.atlassian.net/browse/TNL-8655)

## Testing Instructions

- Checkout to this branch `tinumide/add_post_banners` in this repository 
- Login as a Discussions Moderator user
- To create a Discussions Moderator user visit the instructor tab as staff here: http://localhost:18000/courses/course-v1:edX+DemoX+Demo_Course/instructor#view-membership (for the demo course) and in the membership tab, add the user to a role under the "Course Team Management" section.
- Verify that banners are on posts that have been `Pinned`, `Endorsed`, `Reported` or `Closed` as seen in the mockup [here](https://www.figma.com/file/LVFnO9oRHYiamoicZuXDAC/Discussion-page?node-id=1840%3A35281)
- Login as a Learner
- Verify that banners are on posts that have been `Pinned`


edx-webhook:
Thanks for the pull request, @tinumide! I've created [BLENDED-964](https://openedx.atlassian.net/browse/BLENDED-964) to keep track of it in Jira.   More details are on the [BD-38](https://openedx.atlassian.net/wiki/spaces/COMM/pages/2790490201/BD-38+Discussions+In-Context+Discussion+Moderation+Phase+2) project page.


src/discussions/comments/CommentsView.jsx
```diff
@@ -41,13 +42,16 @@ function CommentsView({ intl }) {
...
           <div className="list-group">
             {comments.map(reply => (
-              <div key={reply.id} className="list-group-item list-group-item-action">
-                <Reply reply={reply} />
-              </div>
+              <><PostBanner post={reply} intl={intl} />
+                <div key={reply.id} className="list-group-item list-group-item-action">
+                  <Reply reply={reply} />
+                </div>
+              </>
```
xitij2000, line 54: Putting the banner like this means you need to move the key to the empty fragment.

src/discussions/comments/data/api.js
```diff
@@ -78,11 +78,13 @@ export async function postComment(comment, threadId, parentId) {
...
   const url = `${commentsApiUrl}${commentId}/`;
-
+  const patchData = snakeCaseObject({
+    voted,
+  });
   const { data } = await getAuthenticatedHttpClient()
-    .patch(url, { raw_body: comment }, { headers: { 'Content-Type': 'application/merge-patch+json' } });
+    .patch(url, patchData, { headers: { 'Content-Type': 'application/merge-patch+json' } });
   return data;
 }
```
xitij2000, line 89: This will break existing code, since you'll no longer be able to update the comment, just mark it as voted/unvoted. 
Remove all data changes from this since this is just a presentational change.

src/discussions/comments/data/slices.js
```diff
@@ -3,6 +3,13 @@ import { createSlice } from '@reduxjs/toolkit';
 
 import { RequestStatus } from '../../../data/constants';
 
+function normaliseProfileImage(currentComment, newComment) {
+  newComment.authorAvatars = newComment.users
+    ? newComment.users?.[newComment.author].profile.image
+    : currentComment.authorAvatars;
+  return newComment;
+}
+
```
xitij2000, line 12: This seems to be part of a prior code, Please remove it.

src/discussions/comments/reply/Reply.jsx
```diff
@@ -17,7 +17,7 @@ function ReplyHeader({ reply, intl }) {
   return (
     <div className="d-flex flex-row justify-content-between">
       <div className="align-items-center d-flex flex-row">
-        <Avatar className="m-2" alt={reply.author} src={reply.users[reply.author].profile.image.image_url_small} />
+        <Avatar className="m-2" alt={reply.author} src={reply.author.Avatar} />
         <div className="status small">
           <a href="#nowhere">
             <h1 className="font-weight-normal text-info-300 mr-1 small">
```
xitij2000, line 23: Please remove this unrelated change, there is another active PR fixing things here.

src/discussions/posts/post/PostBanner.jsx
```diff
@@ -0,0 +1,74 @@
...
+  return null;
+}
+
+function PostBanner({ post, intl }) {
+  const { authenticatedUser } = useContext(AppContext);
+  const bannerProps = banner(post);
+  // show pinned, reported, endorsed and closed posts for moderators
+  if (authenticatedUser.administrator && bannerProps) {
+    return (
+      <div>
```
xitij2000, line 39: Extra div isn't needed. Please remove.

src/discussions/posts/post/PostBanner.jsx
```diff
@@ -0,0 +1,74 @@
...
+    return { variant: 'info', icon: Pin, message: messages.pinned };
+  }
+  return null;
+}
+
+function PostBanner({ post, intl }) {
+  const { authenticatedUser } = useContext(AppContext);
+  const bannerProps = banner(post);
+  // show pinned, reported, endorsed and closed posts for moderators
+  if (authenticatedUser.administrator && bannerProps) {
```
xitij2000, line 37: You don't need to check this, if the props exist, render the component. They should only exist for the correct type of user.

src/discussions/posts/post/PostBanner.jsx
```diff
@@ -0,0 +1,74 @@
...
+    return (
+      <div>
+        <Alert
+          className="d-flex flex-row align-items-start mb-0 p-3"
+          variant={bannerProps.variant}
+          icon={bannerProps.icon}
+        >
+          <>{intl.formatMessage(bannerProps.message)}</>
```
xitij2000, line 45: I think you should let the parent component pass in this message since it's more related to the parent.

src/discussions/posts/post/PostBanner.jsx
```diff
@@ -0,0 +1,74 @@
...
+      </div>
+    );
+  }
+
+  return null;
+}
+
+PostBanner.propTypes = {
+  // eslint-disable-next-line react/require-default-props
+  post: postShape || commentShape,
```
xitij2000, line 70: I don't think this is allowed for proptypes? I think you need to use PropTypes.oneOf

src/discussions/posts/post/PostBanner.jsx
```diff
@@ -0,0 +1,74 @@
...
+
+import messages from './messages';
+import { postShape } from './Post';
+import { commentShape } from '../../comments/comment/Comment';
+
+function banner(post) {
+  if (post.abuseFlagged) {
+    return { variant: 'danger', icon: Flag, message: messages.contentReported };
+  }
+  if (post.hasClosed) {
```
xitij2000, line 21:
```suggestion
  if (post.closed) {
```

xitij2000 (changes requested):
I couple of things:
- I don't think this really needs a separate component. This is used in three places and the styling is different across all of them so using a shared component is an overkill. 
- None of the styles match that well currently. I think the padding is off and for the pinned post the colours are off. 
- The closed banner doesn't show at all. 
- Don't make any changes to the data layer. For now you can use the old UI to add and modify content. 

Please rebase on the latest code.

src/components/Banner.jsx
```diff
@@ -0,0 +1,19 @@
...
+import { Alert } from '@edx/paragon';
+import {
+  Flag,
+  MoreVert,
+  Pin,
+  VerifiedBadge,
+} from '@edx/paragon/icons';
+
+const BANNER_TYPE = {
+  closed: (message) => <Alert className="d-flex flex-row align-items-start mb-0 p-2 border border-radius-1" variant="gray" icon={MoreVert}><>{message}</></Alert>,
```
tinumide, line 12: The color of the banners for closed posts and pinned posts is `light-200` but in use it seems to be way lighter than what is in the mockup.

xitij2000: The mockups often use a slightly different theme, so some differences are to be expected. Go for what the mockup suggests. 

The icon here is wrong, though, it should be `QuestionAnswer` for closed. 

Also try to keep as few styles here as possible. I think border and border radius can be avoided here. 

Here is my implementation of "endorsed" for example:

```jsx
<Alert variant="success" className="p-2 m-0 rounded-0" icon={VerifiedBadge}>
  {intl.formatMessage(messages.endorsedLabel, { endorsersRole: comment.endorsedByLabel })}
</Alert>
```

Finally, I see you've used empty brackets (`<> </>`) in a lot of places where they aren't needed.  You can think of empty brackets as an alternative to array brackets `[]`. If you have a situation where you return multiple items from some component or function they can be useful. Here is better info on them: https://reactjs.org/docs/fragments.html

src/discussions/comments/CommentsView.jsx
```diff
@@ -41,12 +45,19 @@ function CommentsView({ intl }) {
   return (
     <div className="discussion-comments d-flex flex-column w-100 ml-3">
       <div className="mb-2">
+        { thread.closed && BANNER_TYPE.closed(intl.formatMessage(bannerMessages.closed)) }
+        { thread.abuseFlagged && BANNER_TYPE.reported(intl.formatMessage(bannerMessages.contentReported)) }
+        { thread.pinned && BANNER_TYPE.pinned(intl.formatMessage(bannerMessages.pinned)) }
```
xitij2000, line 50: These banners are in the wrong place. 
The only banner I see in the mockups is for closed threads. If a thread is pinned or reported, it gets a banner in the preview on the left side.

src/discussions/comments/CommentsView.jsx
```diff
@@ -41,12 +45,19 @@ function CommentsView({ intl }) {
...
         <div className="list-group list-group-flush">
           <Post post={thread} />
           <div className="list-group">
             {comments.map(comment => (
-              <div key={comment.id} className="list-group-item list-group-item-action">
-                <Comment comment={comment} />
+              <div key={comment.id}>
+                { comment.endorsed && BANNER_TYPE.endorsed(intl.formatMessage(bannerMessages.endorsed)) }
+                { comment.abuseFlagged && BANNER_TYPE.reported(intl.formatMessage(bannerMessages.contentReported)) }
```
xitij2000, line 57: We can just move this inside the Comment component? Then they will show up for both comments here and replies to comments. 
Currently, the banner doesn't show for replies.

src/discussions/posts/post/PostLink.jsx
```diff
@@ -31,12 +34,9 @@ function PostLink({
...
     >
-      {post.abuseFlagged && (
-        <div className="align-items-center bg-danger-100 d-flex flex-fill p-1">
-          <Icon className="text-danger-700" src={Flag} />
-          <span className="text-gray-700 x-small">{intl.formatMessage(messages.contentReported)}</span>
-        </div>
-      )}
+      { post.closed && BANNER_TYPE.closed(intl.formatMessage(messages.closed))}
+      { post.abuseFlagged && BANNER_TYPE.reported(intl.formatMessage(messages.contentReported)) }
+      { post.pinned && BANNER_TYPE.pinned(intl.formatMessage(messages.pinned)) }
```
xitij2000, line 39: I don't think you need to add the post closed banner here. 
Also move these banners inside the Post component and you won't need to repeat them twice, just adding them to the Post component will make them show up both on the left and right side when the post is rendered.

src/components/Banner.jsx
xitij2000: I don't really understand why we need to have all banners in one place. Each of these is a dependent on the context It's in, i.e. a comment, a post etc. and can be directly added there. 

They are not heavily reused, so there is nothing to be gained from moving them away from where they are used and in fact it makes dealing with them harder. 

Please move each banner to where it's used along with the message.

xitij2000 (changes requested):
I think you should move the banners into the components that render them. 

You can move the endorsed and reported banners into the <Comment> component. 
The post closed, reported and pinned banners can be moved to the <Post> component.

Note reported is in both, but other than that nothing is used twice. 
The closed banner should not show on the preview on the left side. So you can check if the preview flag is set and not render it. 
It's very unlikely that a post will be both reported and pinned. It's technically possible, yes, unlikely to happen in practice. 

I think once things are moved back in place and match up. This should be good to go.

xitij2000:
The mockups have changed sufficiently that this is no longer applicable.

edx-webhook:
@tinumide Even though your pull request wasn’t merged, please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/frontend-app-discussions&entry.867055334=https://github.com/edx/frontend-app-discussions/pull/21&entry.1484655318=https://github.com/tinumide&entry.752974735=2021-09-16+03:44&entry.1917517419=2021-09-22+13:14&entry.2133058324=No) so we can improve your experience in the future.

PR #23, opened 2021-09-28
feat: add pagination for responses to comments [BD-38] [TNL-8657] [BB-4735] [WIP]

Cup0fCoffee:


edx-webhook:
Thanks for the pull request, @Cup0fCoffee! I've created [BLENDED-970](https://openedx.atlassian.net/browse/BLENDED-970) to keep track of it in Jira.   More details are on the [BD-38](https://openedx.atlassian.net/wiki/spaces/COMM/pages/2790490201/BD-38+Discussions+In-Context+Discussion+Moderation+Phase+2) project page.


This is currently a draft pull request. When it is ready for our review and all tests are green, click "Ready for Review", or remove "WIP" from the title, as appropriate.

src/discussions/comments/data/__factories__/comments.factory.js
```diff
@@ -51,3 +51,25 @@ Factory.define('commentsResult')
...
+      next,
+      prev,
+      count,
+      num_pages: numPages,
+    };
+  })
+  .attr('results', ['count', 'pageSize', 'page', 'parentId', 'threadId'], (count, pageSize, page, parentId, threadId) => {
+    const len = (pageSize * page <= count) ? pageSize : count % pageSize;
+    return Factory.buildList('comment', len, { thread_id: threadId, parent_id: parentId });
+  });
```
xitij2000, line 75: I don't think you really need this, it's the same as "commentsResult" with the parent id set.

src/discussions/comments/CommentsView.test.jsx
```diff
@@ -142,58 +173,105 @@ describe('CommentsView', () => {
...
-    await screen.findByText(lastPageComment.renderedBody);
+    await screen.findByText(lastPageComment.rendered_body);
     await expect(findLoadMoreCommentsButton()).rejects.toThrow();
   });
+
+  describe('comments responses', () => {
+    it('initially loads only the first page', async () => {
+      renderComponent();
+
+      await screen.findByText('comment number 1', { exact: false });
```
xitij2000, line 239: I think this should be wrapped in an expect as well.

xitij2000 (approved):
:+1: Looks good, but I'm seeing an issue when adding comments. It causes the comments to be appended in front of existing comments. 

When a new comment is added to a comment, we refresh the comment list, this is causing issues. 

I think the solution here is to simply not refresh.

- [x] I tested this: tested UI manually
- [x] I read through the code
- [na] Includes documentation

src/discussions/comments/data/__factories__/comments.factory.js
Cup0fCoffee: Did see that at the time I was writing it, but you are right, thank you for spotting it.
I will remove it.

src/discussions/comments/CommentsView.test.jsx
Cup0fCoffee: @xitij2000 Can you please explain why?
If the element has been found, expect doesn't add anything.
If the element hasn't been found, an error would be thrown, which would lead to the test failing and the correct message would be displayed, and expect is not needed for that.
If there was an element, let's say a loader, which I could wait for to disappear, it would be easy to change `findByText` to `queryByText` and wrap it with a expect. If you think it's worth it, I can change the code to wait for `store.comments.status` to change to `RequestStatus.SUCCESSFUL` and then do the assertion, but my understanding was that it's bad to rely on internal state, and instead try to rely on UI.

src/discussions/comments/data/slices.js
```diff
@@ -20,26 +20,31 @@ const commentsSlice = createSlice({
...
+    responsesPagination: {},
   },
   reducers: {
     fetchCommentsRequest: (state) => {
       state.status = RequestStatus.IN_PROGRESS;
     },
     fetchCommentsSuccess: (state, { payload }) => {
+      const { threadId } = payload;
+      // force endorsed to be null, true or false
+      const endorsed = payload.endorsed === undefined ? null : payload.endorsed;
```
mehaknasir, line 33: why are we assigning null to `endorsed`

src/discussions/comments/data/slices.js
```diff
@@ -20,26 +20,31 @@ const commentsSlice = createSlice({
...
       state.status = RequestStatus.SUCCESSFUL;
-      state.commentsInThreads[payload.threadId] = [
-        ...(state.commentsInThreads[payload.threadId] || []),
-        ...(payload.commentsInThreads[payload.threadId] || []),
+      state.commentsInThreads[threadId] = state.commentsInThreads[threadId] ?? {};
+      state.pagination[threadId] = state.pagination[threadId] ?? {};
+      state.commentsInThreads[threadId][endorsed] = [
+        ...(state.commentsInThreads[threadId][endorsed] || []),
+        ...(payload.commentsInThreads[threadId] || []),
```
mehaknasir, line 39: when you assigned null to endorsed and then you assign values here it makes a weird redux structure like this
Screenshot 2021-09-30 at 7 16 46 PM


I dont understand the logic here? dont we have some other solution to differentiate between endorsed and unendorsed comments?

Cup0fCoffee: @mehaknasir 
@xitij2000 can correct me if I'm wrong, but I think the idea here is such: questions have endorsed and unendorsed comments, and discussions don't have such differentiation. So when the thread (aka post) is a question, `endorsed` can have values `true` and `false`, and for threads that are discussions we need to set `endorsed` to a value which signifies the absence of differentiation, and since `null` signifies the absence of value, I think it's suitable to use it in this manner.
I'm sure there are alternative approaches, and I also think current solution is one of the cases where `null` is appropriately used for semantics.

Cup0fCoffee: Btw, I need to mention, that these changes are from #22, I just rebased my branch onto it, because there are some changes that this PR depends on, so may be it's more appropriate to discuss it there.

src/discussions/comments/CommentsView.test.jsx
xitij2000: Fair enough!

src/discussions/comments/data/slices.js
xitij2000: There are two types of posts, discussions and questions. 
Discussions are simple, and only track a single type of comment. 
Questions are more complex. They differentiate between endorsed and unendorsed comments. For questions, an endorsed comment is paginated separately, and displayed separately. 

So we need to track comments on the basis of endorsement: null if they are on a discussion, true/false if they on a question post and are endorsed/not-endorsed.

src/discussions/comments/data/slices.js
xitij2000: We can as easily create a mapping where we use:
"endorsed" instead of true
"unendorsed" instead of false
"discussion" instead of null

However, since the redux state is entirely internal and not seen by users I didn't see much reason for adding this complexity.

mehaknasir: Even if its related to internal implementation I would still like to have some meaningful structuring of redux data. for maintaining code readability and quality. I would suggest changing this.

xitij2000: I'm afraid I disagree here. It's the code that should be readable here, not the data. We're dealing with raw ids like "906e91ed8a464499aef4561829", none of that is supposed to be readble. It is supposed to be structured in a way to allow easy access to data. That is why the data is being restructured from the original API response to something that makes queries more efficient. 

I don't understand how the code will become more readable, or how its quality will improve if instead of a simple operation like:

``commentsInThreads[threadId][endorsed] ``

we now have:

``commentsInThreads[threadId][ENDORSED_NAME_MAP[endorsed]]``

To me, the new code above is a lot **less** readable (whatever you call the variable instead of `ENDORSED_NAME_MAP`). Every time we need to fetch or save this data we'll have to map this value.

awais-ansari: @xitij2000 PR is looking good. I totally agree with @mehaknasir suggestion. Assigning a readable name instead of null will increase more readable structure. We also posted an open suggestion question related to this practice in fedx. We will approve this PR soon.

kenclary: (please excuse the intrusion)

Under the current suggestions, you would still use `commentsInThreads[threadId][endorsed]`, however, `endorsed` would generally be a string value, either `'endorsed'`, `'unendorsed'`, or `'discussion'`. As a result, the object keys will be human readable.

xitij2000: > Under the current suggestions, you would still use commentsInThreads[threadId][endorsed], however, endorsed would generally be a string value, either 'endorsed', 'unendorsed', or 'discussion'.
@kenclary 
Here is the thing, endorsed is a value returned from the API, its value is either `true`, or `false`. Its value isn't "endorsed" or "unendorsed", so we'll need to make the mapping at some point. I just don't see the reason why. 

This is a data structure that is only ever accessed by the code, humans do not look at it unless they install a specific browser extension and use it to inspect the state when the code is running in a development environment. 

I really don't understand the need for human-readable data structures here. We don't inspect the value of each python variable in code to make sure that it's contents are human-readable. We see if the code is readable, and if the data is structured in a way that allows efficient access. 

I could understand the argument if this was supposed to be saved to a file and read by humans at any point. I mean, we can go further here and use the title of the post instead of its id, as that will be more human-readable. 

I don't want to endlessly block this work, so I'll update my PR, but I don't think human readability should ever be a concern when creating temporary data structures that will never be seen by a human outside of debugging. If this is to become a recommended frontend or backend practice then I would strongly urge against it.

xitij2000: I've updated the code here: https://github.com/edx/frontend-app-discussions/pull/22/commits/07dc355e231c7397a84a81304004c0a5cbcde210 
@Cup0fCoffee Could you rebase on my latest changes once they are approved.

src/discussions/comments/data/slices.js
```diff
@@ -41,7 +42,7 @@ const commentsSlice = createSlice({
         totalPages: payload.pagination.numPages,
         hasMorePages: Boolean(payload.pagination.next),
       };
-      state.commentsInComments = { ...state.commentsInComments, ...payload.commentsInComments };
+      // state.commentsInComments = { ...state.commentsInComments, ...payload.commentsInComments };
```
xitij2000, line 45: I guess this can just be removed now?

xitij2000:
> @xitij2000, the issue with adding a response when there are more responses to load still persists.
> I removed the code which fetches comments when you add a new one, as you suggested. It is better like that, but we still end up with a new comment in-between old ones.
> I'm writing this comment mostly to document the issue, since it doesn't seem like we are going to fix it right now. If you want, I can add it to the commit message.

I think another way would be to maintain a separate list of newly-added comments, so the UI will be able to render this comment at the end, after the load-more-responses button. When we see the id of the newly-added comment in the list comment list, we can remove it from that list. 

It's a larger change that's not in scope here, so you can create a new ticket for this and we can look into it later.

src/discussions/comments/data/slices.js
Cup0fCoffee: You're right, I will remove it.
Thank you for pointing it out.

Cup0fCoffee:
> It's a larger change that's not in scope here, so you can create a new ticket for this and we can look into it later.

Created ticket for it - [BB-5007](https://tasks.opencraft.com/browse/BB-5007).

PR #34, opened 2021-11-03
build: use shared browserslist config

adamstankiewicz:
* Removes custom `browserslist` configuration in favor of using a [shared configuration](https://github.com/edx/browserslist-config).
* Removes `is-es5` check in CI since ES5 was only needed for IE 11 support which is dropped. The supported browsers defined by the shared configuration all [natively support ES6](https://caniuse.com/?search=es6).

This change reduces the resultant asset bundle size.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,5 @@ jobs:
       run: npm run build
     - name: i18n_extract
       run: npm run i18n_extract
-    - name: is-es5
-      run: npm run is-es5
     - name: Coverage
       uses: codecov/codecov-action@v2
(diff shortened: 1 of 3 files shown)


xitij2000:
@adamstankiewicz I looked into the build failures here a bit and I think upgrading to node v16 should solve them. 
Node v12 doesn't support some of the ES6 features used by the code here, so by targeting newer versions of browsers, this is also targeting a version of JS newer than what node itself supports. 

To avoid this issue, perhaps the browser list config should include the current version of node being used? The tests all run in Node so they need to be able to run the code as well.

xitij2000:
@adamstankiewicz If you rebase on the latest changes, the tests should start passing.

mehaknasir (approved):

A-ASAD (approved):

PR #37, opened 2021-11-19
feat: Make UI more responsive to smaller screens and embeds [BD-38] [TNL-8430] [BB-4420]

xitij2000:
Updates the UX so that on a smaller screen either the topic/post sidebar is displayed or the post view to give each more space.

Also adds support for an in-context view where the header and navigation is replaced with a sidebar-specific header.

Mobile UI:

https://user-images.githubusercontent.com/118837/142617888-d9562e00-271c-4c81-9c08-aeefad3610d3.mp4

In-Context UI:

https://user-images.githubusercontent.com/118837/142617860-53663a18-3448-4484-8dd3-17b53b2ef591.mp4

## Testing

- With different pages open, try using the UI with the browser resized down to a mobile size, on a mobile, or using browser tools to test mobile UIs
- Append `?inContext` at the end of the usr and try the same. This should display the in-context UI which is designed to be loaded in a sidebar in the learning MFE and as such will hide the header, footer and change some UI elements.


edx-webhook:
Thanks for the pull request, @xitij2000! I've created [BLENDED-1019](https://openedx.atlassian.net/browse/BLENDED-1019) to keep track of it in Jira.   More details are on the [BD-38](https://openedx.atlassian.net/wiki/spaces/COMM/pages/2790490201/BD-38+Discussions+In-Context+Discussion+Moderation+Phase+2) project page.


src/discussions/posts/post-actions-bar/PostActionsBar.jsx
```diff
@@ -1,36 +1,63 @@
...
+        className="ml-2 my-0"
         onClick={() => dispatch(showPostEditor())}
       >
         {intl.formatMessage(messages.addAPost)}
       </Button>
+      {inContext && (
+      <>
+        <div className="border-right mr-3 ml-4" />
+        <IconButton src={Close} iconAs={Icon} onClick={handleCloseInContext} alt={intl.formatMessage(messages.close)} />
+      </>
```
Cup0fCoffee, line 51: nit: should this be indented one level more?

src/discussions/posts/post-editor/PostEditor.jsx
```diff
@@ -229,22 +244,22 @@ function PostEditor({
             </Form.Group>
             {canSelectCohort
               && (
```
Cup0fCoffee, line 246: nit:
```suggestion
            {canSelectCohort && (
```

xitij2000: I think the current style is consistent with the rest of the usages.

package.json
```diff
@@ -58,15 +58,15 @@
     "regenerator-runtime": "0.13.9",
     "timeago.js": "4.0.2",
     "tinymce": "5.10.2",
-    "yup": "0.32.11"
+    "yup": "0.31.1"
```
Cup0fCoffee, line 61: [Downgraded](https://github.com/edx/frontend-app-discussions/pull/37/commits/05db3a168d64b33d00497bff35d96de16c15a67b#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R61)?

src/discussions/posts/post-editor/PostEditor.jsx
```diff
@@ -77,19 +79,30 @@ function PostEditor({
...
+    anonymous: false,
+    anonymousToPeers: false,
   };
+  if (editExisting) {
+    initialValues = {
+      postType: post.type || 'discussion',
+      topic: post.topicId || topicId || nonCoursewareTopics?.[0]?.id,
+      title: post.title || '',
+      comment: post.rawBody || '',
+      follow: (post.following === null || post.following === undefined) ? true : post.following,
```
Cup0fCoffee, line 97: Not sure why we need to check for both `null` and `undefined`. I understand that we can't do `!post.following`, because it can be `false`, and we are trying to [catch "nullish" values](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator). Would that be better, since it's a more concise equivalent:
```suggestion
      follow: post.following ?? true,
```

src/discussions/posts/post-editor/PostEditor.jsx
```diff
@@ -109,7 +122,7 @@ function PostEditor({
     } else {
       const cohort = canSelectCohort
         // null stands for no cohort restriction ("All learners" option)
-        ? (values.cohort ?? null)
+        ? (values.cohort || null)
```
Cup0fCoffee, line 125: I saw that there was a commit that replaced `??` with `||`, and I didn't pay much attention to it. I understand that here it doesn't really make any difference, since `values.cohort` is either a non empty string or a "nullish" value, but `??` semantically hints that we are checking for "nullish" values.
Can you help me understand what is the purpose of this change?

src/discussions/posts/post-actions-bar/PostActionsBar.jsx
```diff
@@ -45,10 +45,15 @@ function PostActionsBar({
...
-      <>
-        <div className="border-right mr-3 ml-4" />
-        <IconButton src={Close} iconAs={Icon} onClick={handleCloseInContext} alt={intl.formatMessage(messages.close)} />
-      </>
+        <>
+          <div className="border-right mr-3 ml-4" />
+          <IconButton
+            src={Close}
+            iconAs={Icon}
+            onClick={handleCloseInContext}
```
Cup0fCoffee, line 53: Is the implementation of `handleCloseInContext` intended to be done in the future?

Cup0fCoffee:
:+1: Conditionally approved, as long as the feedback is addressed and rebased onto master, as right now the PR contains commits from another PR that has been merged.

- [x] I tested this
- [x] I read through the code

package.json
xitij2000: Yes, it seems that the newer version of yup causes certain lint checks to fail. We can upgrade once the code has moved to Node 16 and support for IE11 has been dropped.

src/discussions/posts/post-editor/PostEditor.jsx
xitij2000: The purpose of this change is that in some cases this is causing the tests to fail because the tests run on node 12, and node 12 doesn't support `??` yet. I don't know why this is happening, but tests consistently fail in this case. 

I've replaced situations where we're dealing with non booleans from ?? to || since it shouldn't make a different.

src/discussions/posts/post-actions-bar/PostActionsBar.jsx
xitij2000: Yes, because this is intended to be embedded in an iframe in a sidebar and for this to close the sidebar we need to implement a postMessage-based API that will request the embedding app to close the sidebar.

src/discussions/navigation/breadcrumb-menu/BreadcrumbDropdown.jsx
```diff
@@ -0,0 +1,60 @@
...
+  // eslint-disable-next-line react/forbid-prop-types
+  currentItem: PropTypes.object.isRequired,
+  intl: intlShape.isRequired,
+  showAllPath: PropTypes.func.isRequired,
+  // eslint-disable-next-line react/forbid-prop-types
+  items: PropTypes.array.isRequired,
+  itemPathFunc: PropTypes.func.isRequired,
+  itemLabelFunc: PropTypes.func.isRequired,
+  itemActiveFunc: PropTypes.func.isRequired,
+};
```
mehaknasir, line 58: all are required but I cant see any default declaration for required props

xitij2000: @mehaknasir  required props don't need a default declaration, because they are always provided, only non-required props need defaults.

src/discussions/posts/post-editor/PostEditor.jsx
xitij2000: As I mentioned elsewhere currently ?? is breaking tests and lint checks in some cases, which is why I've switched it out. 

In other cases ?? is equivalent to || but not here, so I need to check for null or undefined.

src/discussions/navigation/breadcrumb-menu/BreadcrumbDropdown.jsx
mehaknasir: I recently came across a project in which it required me to add the default of required props which made me confused. but I just remember we are adding defaults for non required. resolving it now.

mehaknasir (approved):
The changes looks good to me. Approving

edx-webhook:
@xitij2000 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/frontend-app-discussions&entry.867055334=https://github.com/edx/frontend-app-discussions/pull/37&entry.1484655318=https://github.com/xitij2000&entry.752974735=2021-11-19+11:47&entry.1917517419=2021-12-02+12:24&entry.2133058324=Yes) so we can improve your experience in the future.

PR #55, opened 2022-02-03
feat: integrate feeder package

awais-ansari:
[TNL-9509](https://openedx.atlassian.net/browse/TNL-9509)

- Integrate [feeder](https://feeder.sh/dashboard) package
- Email field will not visible to the users.
- The API payload will automatically carry user email.
- Feedback message is required.
- 
**Note**
- FEEDER_PROJECT_ID is required to enable the feedback module on stage and production.

Screenshot 2022-02-03 at 10 47 32 PM
Screenshot 2022-02-03 at 10 48 05 PM
Screenshot 2022-02-03 at 10 49 08 PM
Screenshot 2022-02-03 at 10 49 30 PM


.env.development
```diff
@@ -20,3 +20,4 @@ REFRESH_ACCESS_TOKEN_ENDPOINT='http://localhost:18000/login_refresh'
 SEGMENT_KEY=''
 SITE_NAME=localhost
 USER_INFO_COOKIE_NAME='edx-user-info'
+FEEDER_PROJECT_ID='61f123796593f20004018b2f'
```
awaisdar001, line 23: you shouldn't put API keys here, these should be carried over using env settings in edx internals for MFEs. please remove these.

src/discussions/discussions-home/DiscussionsHome.test.jsx
```diff
@@ -29,10 +29,16 @@ function renderComponent(location = `/${courseId}/`) {
 
 describe('DiscussionsHome', () => {
   beforeEach(async () => {
+    setConfig({
+      ...getConfig(),
+      FEEDER_PROJECT_ID: 'abc123',
```
awaisdar001, line 34: is that a test ID? `abc123` => `test-id` would be more readable.

awaisdar001 (changes requested):
Great job for integrating the feedback tool in such a tight timeline.

Code quality is excellent 👍

src/components/Feedback/constants.js
```diff
@@ -0,0 +1,2 @@
+export const FeedbackTypes = ['Review', 'Bug', 'Idea'];
+export const primaryColor = '#00262b';
```
mehaknasir, line 2: do we have this color code available in paragon? can we use that from there?

.env.development
awais-ansari: Sure. I'll remove it from this MEF.

src/discussions/discussions-home/DiscussionsHome.test.jsx
awais-ansari: Yes. It is a test ID. I'll update it.

src/components/Feedback/constants.js
awais-ansari: We can only pass the color CSS value such as HEX, RGB, and paragon provides string classes such as `text-primary-500` which is not compatible with it. I tried this solution too.

awaisdar001 (approved):

mehaknasir (approved):

PR #58, opened 2022-02-08
fix: remove app header and footer

awais-ansari:
- Remove app header and footer for both in-context or non-in-context discussions

**Description**
While rendering the mfe in an iframe within discussions tab, there is a header and footer rendered as well which seems duplicate as discussion tab already has its own header and footer. 

This PR removes the header and footer so that the rendering within the iframe looks more closer to the mockups. 

Screenshot 2022-02-08 at 12 57 25 PM

FYI: @xitij2000

### DIFF
--- a/src/discussions/discussions-home/DiscussionsHome.jsx
+++ b/src/discussions/discussions-home/DiscussionsHome.jsx
@@ -3,11 +3,14 @@ import React, { useContext, useEffect } from 'react';
 import classNames from 'classnames';
 import { useDispatch, useSelector } from 'react-redux';
 import {
-  Redirect, Route, Switch, useHistory, useLocation, useRouteMatch,
+  Redirect,
+  Route,
+  Switch,
+  useHistory,
+  useLocation,
+  useRouteMatch,
 } from 'react-router';
 
-import Footer from '@edx/frontend-component-footer';
-import Header from '@edx/frontend-component-header';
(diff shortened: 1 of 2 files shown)


AhtishamShahid (approved):

awaisdar001 (approved):

xitij2000:
@awais-ansari This also mean there will be no header and footer when the MFE is viewed as a non-embedded app. Is that OK? I had a ticket in the upcoming sprint to address this by removing the header and footer only when the app is embedded in an iframe. 

If the intention is to not expose the MFE directly then this approach makes sense. In that case we can even remove the dependency on these header files and make the MFE lighter.

awaisdar001:
thank you for the context @xitij2000 . I am not aware of any use case where the MFE needs to be rendered without the embedded iframe. but I can as for more info from product. 

:+1 for removing the dependency of header and footer.

awais-ansari:
> @awais-ansari This also mean there will be no header and footer when the MFE is viewed as a non-embedded app. Is that OK? I had a ticket in the upcoming sprint to address this by removing the header and footer only when the app is embedded in an iframe.
> 
> If the intention is to not expose the MFE directly then this approach makes sense. In that case we can even remove the dependency on these header files and make the MFE lighter.

Thank you for your input. I totally agreed on removing the dependency as currently, we do not have any use case where the MFE renders without an iframe.  I'll raise a PR to remove the dependency.

PR #64, opened 2022-02-22
feat: independent scrolling for posts list and posts [BD-38] [TNL-9571] 

xitij2000:
Adds independent scrolling for left and right panels.

### DIFF
--- a/src/discussions/discussions-home/DiscussionsHome.jsx
+++ b/src/discussions/discussions-home/DiscussionsHome.jsx
@@ -78,8 +78,8 @@ export default function DiscussionsHome() {
         inContext,
       }}
     >
-      <main className="container-fluid d-flex flex-column p-0">
-        <div className="d-flex flex-row justify-content-between shadow navbar">
+      <main className="container-fluid d-flex flex-column p-0 h-100 w-100 overflow-hidden">
+        <div className="d-flex flex-row justify-content-between shadow navbar fixed-top">
           {!inContext && (
             <Route path={Routes.DISCUSSIONS.PATH} component={NavigationBar} />
           )}
@@ -89,9 +89,9 @@ export default function DiscussionsHome() {
           path={[Routes.POSTS.PATH, Routes.TOPICS.CATEGORY]}
           component={LegacyBreadcrumbMenu}
         />
-        <div className="d-flex flex-row">
+        <div className="d-flex flex-row overflow-hidden">
           <div
-            className={classNames('flex-column w-25 w-xs-100 w-lg-25', {
+            className={classNames('flex-column w-25 w-xs-100 w-lg-25 overflow-auto h-100 pb-2', {
               'd-none': !displaySidebar,
               'd-flex': displaySidebar,
             })}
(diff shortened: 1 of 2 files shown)


edx-webhook:
Thanks for the pull request, @xitij2000! I've created [BLENDED-1107](https://openedx.atlassian.net/browse/BLENDED-1107) to keep track of it in Jira.   More details are on the [BD-38](https://openedx.atlassian.net/wiki/spaces/COMM/pages/2790490201/BD-38+Discussions+In-Context+Discussion+Moderation+Phase+2) project page.


src/discussions/discussions-home/DiscussionsHome.jsx
```diff
@@ -78,8 +78,8 @@ export default function DiscussionsHome() {
         inContext,
       }}
     >
-      <main className="container-fluid d-flex flex-column p-0">
-        <div className="d-flex flex-row justify-content-between shadow navbar">
+      <main className="container-fluid d-flex flex-column p-0 h-100 w-100">
+        <div className="d-flex flex-row justify-content-between shadow navbar fixed-top">
```
Cup0fCoffee, line 82: Why do we need the `fixed-top`? Imho, `main` itself should not be scrollable, so the header wouldn't move.

src/discussions/discussions-home/DiscussionsHome.jsx
```diff
@@ -89,9 +89,9 @@ export default function DiscussionsHome() {
           path={[Routes.POSTS.PATH, Routes.TOPICS.CATEGORY]}
           component={LegacyBreadcrumbMenu}
         />
-        <div className="d-flex flex-row">
+        <div className="d-flex flex-row vh-100">
```
Cup0fCoffee, line 92: I think `vh-100` is usually used when the container should be the same height as the view port, which is not the case here. We want the container to take the rest of the space in the flexbox wrapper, so `flex-grow-1` can (I want to say 'should', but I'm not 100% confident) be used here.

Cup0fCoffee: Also, you should add `overflow-hidden` on `main` and this container, so the children can be scrollable. [Ref](https://stackoverflow.com/a/35609992/15935523)

Cup0fCoffee (changes requested):
I think we should utilize the fact that `main` is a flexbox, instead of forcing strict height on it's children, since it leads to some bugs.

src/discussions/discussions-home/DiscussionsHome.jsx
xitij2000: Fixed top is a slightly hacky way to get that bar higher in the z-index so that it can show the shadow above the other elements when scrolling.

src/discussions/discussions-home/DiscussionsHome.jsx
xitij2000: I did not know this! Thanks! I've replaced the vh-100 with overflow hidden

AhtishamShahid (approved):

awais-ansari (approved):

edx-webhook:
@xitij2000 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/frontend-app-discussions&entry.867055334=https://github.com/openedx/frontend-app-discussions/pull/64&entry.1484655318=https://github.com/xitij2000&entry.752974735=2022-02-22+13:38&entry.1917517419=2022-02-23+12:36&entry.2133058324=Yes) so we can improve your experience in the future.

PR #65, opened 2022-02-23
feat: show empty state on posts pages [BD-38] [TNL-9484] [SE-5347]

keithgg:
## Description

When a learner is viewing discussion pages and there have not yet been comments show a page noting that there is nothing to show.

## Supporting information

- https://openedx.atlassian.net/browse/TNL-9484
- https://www.figma.com/file/LVFnO9oRHYiamoicZuXDAC/Discussion-page?node-id=4059%3A315300
- https://tasks.opencraft.com/browse/SE-5347

## Testing instructions

Go to the new discussions app

**If you have posts already:**
- You should see the Empty page in the design to the right of the sidebar
- Once you've selected a post, it should disappear.
- The above applies for the My Posts/All Posts/Topics sections.

**If you don't have *any* posts:**
- You should not see the sidebar under My Posts/All Posts. Only the Empty page UI.
- Once a post is added the sidebar should appear again.

## Deadline

8 March 2022

## Other information

- I followed the ADR as far as I'm aware, but please check if done correctly.
- I refactored the DiscussionHome component into two otherwise it would've gotten much bigger with this PR.


edx-webhook:
Thanks for the pull request, @keithgg! I've created [OSPR-6487](https://openedx.atlassian.net/browse/OSPR-6487) to keep track of it in JIRA, where we prioritize reviews. Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket as you can:
- supporting documentation
- [Open edX discussion forum threads](https://discuss.openedx.org/)
- timeline information ("this must be merged by XX date", and why that is)
- partner information ("this is a course on edx.org")
- any other information that can help Product understand the context for the PR

All technical communication about the code itself will be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).


Please let us know once your PR is ready for our review and all tests are green.

xitij2000 (changes requested):
Please update the PR title to include [BD-38] and other relevant ticket ids at the end.

src/discussions/common/EmptyPage.jsx
```diff
@@ -0,0 +1,95 @@
...
+      displayMessage = false;
+      emptyTitle = messages.noPostSelected;
+    }
+  } else if (isOnTopicsPage) {
+    displayMessage = false;
+    emptyTitle = ROUTE_TITLES[match.path];
+  } else if (isFiltered || totalThreads > 0) {
+    displayMessage = false;
+    emptyTitle = ROUTE_TITLES[match.path];
+  }
```
xitij2000, line 65: This seems a bit hacky. I think a better approach would be to not centralise this logic. Have each part of the app show the emptypage component with its own specialised logic and just pass in the empty page message.

xitij2000 (changes requested):
Right now this is just not working for me unfortunately. I just don't see the empty page. 

Also I'm not sure why the svg was replaced with a png?

src/discussions/common/EmptyPage.jsx
keithgg: True. I'll update.

keithgg:
I moved from svg to png, because the svg was causing the tests to fail. 

See the thread here: https://github.com/gregberge/svgr/issues/83

None of the solutions worked for me.

xitij2000:
> I moved from svg to png, because the svg was causing the tests to fail.
> 
> See the thread here: [gregberge/svgr#83](https://github.com/gregberge/svgr/issues/83)
> 
> None of the solutions worked for me.

All the paragon icons we are using are SVG, so there must be some way around this. Could you revert back to svg? I'll have a look and see. 

Perhaps you can create an svg-only component for it, and mock that for tests?

src/discussions/common/EmptyPage.jsx
keithgg: Looks like there isn't a way to completely get rid of this, but it does look cleaner.

src/discussions/discussions-home/DiscussionSidebar.jsx
```diff
@@ -0,0 +1,52 @@
...
+import classNames from 'classnames';
+import {
+  Redirect, Route, Switch, useLocation,
+} from 'react-router';
+
+import { Routes } from '../../data/constants';
+import { PostsView } from '../posts';
+import { TopicsView } from '../topics';
+
+export default function DiscussionSidebar({ displaySidebar }) {
```
xitij2000, line 13: Why not use the hook here?

xitij2000 (changes requested):
This still doesn't work in some of the cases I've tried:
- If you visit an empty topic, it shows "No post selected" instead of "Nothing here yet"
- The small screen view is broken, the sidebar is always visible.

src/discussions/discussions-home/DiscussionSidebar.jsx
keithgg: It's dependent on whether we're displaying the content area or not. On mobile when showing content we hide sidebar.

```javascript
  if (displayContentArea) {
    // If the window is larger than a particular size, show the sidebar for navigating between posts/topics.
    // However, for smaller screens or embeds, only show the sidebar if the content area isn't displayed.
    displaySidebar = windowSize.width >= breakpoints.large.minWidth;
  }
  ```

**Comment by keithgg** [comment]
> If you visit an empty topic, it shows "No post selected" instead of "Nothing here yet"

Sorted. I was looking at the `threads` state instead of just the `topic` state.

File: src/discussions/data/hooks.js [hunk 3]
```diff
@@ -0,0 +1,91 @@
...
+import { fetchCourseBlocks } from '../../data/thunks';
+import { clearRedirect } from '../posts/data';
+import { fetchCourseTopics } from '../topics/data/thunks';
+import { discussionsPath } from '../utils';
+import { selectAreThreadsFiltered } from './selectors';
+import { fetchCourseConfig } from './thunks';
+
+export function usePostThreadCount() {
+  return useSelector((state) => state.threads.totalThreads);
+}
```
xitij2000, line 20: I don't really see the need fof this hook since it's just working as a selector. Instead create a selector and use that with useSelector.

src/discussions/data/hooks.js
```diff
@@ -0,0 +1,91 @@
...
+
+export function useTotalTopicThreadCount() {
+  const topics = useSelector((state) => state.topics.topics);
```
xitij2000, line 23:
```suggestion
  const topics = useSelector(selectTopics);
```

src/discussions/data/hooks.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  if (isViewingTopics) {
+    return true;
+  }
+
+  return totalThreads > 0;
+};
+
+export function useCourseTopicsAndBlocks(courseId) {
```
xitij2000, line 51: This seems to be doing more than just topics and blocks, so maybe call it `useCourseDiscussionData`?

src/discussions/posts/NoResults.jsx
```diff
@@ -0,0 +1,31 @@
+import { useSelector } from 'react-redux';
+
+import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
+
+import { selectAreThreadsFiltered } from '../data/selectors';
+import messages from '../messages';
+
+function NoResults({ intl }) {
```
xitij2000, line 8: This doesn't seem to be used?

xitij2000 (commented):
This is pretty close but I still see a few issues. Once they are fixed this is good to merge.

src/discussions/data/hooks.js
keithgg: Updated.

src/discussions/data/hooks.js
keithgg: Updated.

src/discussions/data/hooks.js
keithgg: Updated.

src/discussions/posts/NoResults.jsx
keithgg: It got removed as part of resolving merging in master.

When filtering the Post views (not Topics) it should show "No Results found" if there isn't anything to show. Re-added.

keithgg:
Thanks for the patience @xitij2000! Changes made as requested. I'll rebase the latest commit once you confirm all good.

xitij2000 (approved):
:+1: 
Tested the following scenarios:
- Tested with empty course, not content in my or all posts
- Tested with course that has content in all posts but no user posts
- Tested with both own and all posts
- Tested without selecting posts
- Tested without selecting topics
- Tested filtering posts returning no results
- Tested on mobile

- [x] I tested this: tested on devstack
- [x] I read through the code

edx-webhook:
@keithgg 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/frontend-app-discussions&entry.867055334=https://github.com/openedx/frontend-app-discussions/pull/65&entry.1484655318=https://github.com/keithgg&entry.752974735=2022-02-23+12:07&entry.1917517419=2022-03-09+12:15&entry.2133058324=Yes) so we can improve your experience in the future.

PR #67, opened 2022-02-24
[TNL-9570] Message icon showing in the absense of comments or responses on the post.

muhammadadeeltajamul:
Added a check while displaying the message icon for comments if comments count is great than 1.

https://openedx.atlassian.net/browse/TNL-9570

### DIFF
--- a/src/discussions/posts/post/PostFooter.jsx
+++ b/src/discussions/posts/post/PostFooter.jsx
@@ -53,7 +53,7 @@ function PostFooter({
             />
           )}
       </OverlayTrigger>
-      {preview
+      {preview && post.commentCount > 1
         && (
           <>
             <Icon src={QuestionAnswer} className="mx-2 my-0" />


awais-ansari (approved):
please add the ticket link in the description.

mehaknasir:
@muhammadadeeltajamul don't we need to show an icon if the count is 1?

src/discussions/posts/post/PostFooter.jsx
```diff
@@ -53,7 +53,7 @@ function PostFooter({
             />
           )}
       </OverlayTrigger>
-      {preview
+      {preview && post.commentCount > 1
```
mehaknasir, line 56: I feel the condition should be `post.commentCount > 0`?

muhammadadeeltajamul: > 
Backend send `post.commentCount` as `1` if the post has no comment.

muhammadadeeltajamul: Backend send `post.commentCount` as `1` if the post has no comment.

mehaknasir: according to adeel the count is coming 1 from backend if there is no comment. it should be fixed from backend no.

muhammadadeeltajamul: It was decided in the meeting that the appropriate way would be to change condition to `post.commentCount > 1` instead of changing from backend.

mehaknasir: okay, I was unaware of that discussion. however, I would like to discuss why we finalized that but approving it for now.

mehaknasir (approved):

PR #71, opened 2022-02-24
[TNL-9572] Removed legacy indicator for unread post

muhammadadeeltajamul:
Removed the Icon component which is rendering when the post is not read

https://openedx.atlassian.net/browse/TNL-9572


awais-ansari (approved):

awaisdar001 (approved):
:shipit:

PR #76, opened 2022-03-01
fix: answered and endorsed indicator modification

A-ASAD:
Ticket: [TNL-9610](https://openedx.atlassian.net/browse/TNL-9610)

- Font of “Answer” and “Endorsed” string is bigger now.
- Color of username is same as rest of the text.
- Added user role (“Staff”, “TA” etc.) and icon.
- Added more space between timestamp and username.
- Top-right and top-left corner of the header box are rounded.

Screenshot 2022-03-01 at 1 39 48 PM
Screenshot 2022-03-01 at 1 39 22 PM

### DIFF
--- a/src/discussions/common/AlertBanner.jsx
+++ b/src/discussions/common/AlertBanner.jsx
@@ -65,7 +65,7 @@ function AlertBanner({
                 />
               )}
 
-              <span className="mr-3">
+              <span className="mr-3" data-testid="endorsed-by-label">
                 {endorsedByLabels[content.endorsedByLabel]}
               </span>
               {timeago.format(content.endorsedAt, intl.locale)}
--- a/src/discussions/common/AlertBanner.jsx
+++ b/src/discussions/common/AlertBanner.jsx
@@ -26,7 +26,12 @@ function AlertBanner({
   return (
     <>
       {content.endorsed && (
-        <Alert variant="plain" className={`p-3 m-0 rounded-0 shadow-none ${classes}`} icon={iconClass}>
+        <Alert
+          variant="plain"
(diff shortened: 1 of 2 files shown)


awais-ansari (approved):

src/discussions/common/AlertBanner.jsx
```diff
@@ -20,24 +22,52 @@ function AlertBanner({
...
+              <span className="mx-2">{content.endorsedBy}</span>
+
+              {content.endorsedByLabel === 'Staff' ? (
+                <Icon
+                  style={{
+                    width: '1rem',
+                    height: '1rem',
+                  }}
+                  src={Institution}
+                />
```
mehaknasir, line 57: can we add a test case for this condition as well? it will remove the codecov failure.

A-ASAD: Sure

src/discussions/common/AlertBanner.test.jsx
```diff
@@ -0,0 +1,66 @@
...
+function renderComponent(
+  content,
+) {
+  render(
+    <IntlProvider locale="en">
+      <AppProvider store={store}>
+        <AlertBanner
+          content={content}
+          postType={ThreadType.QUESTION}
+        />
```
mehaknasir, line 31: we are looking for word TA and staff in the alert banner anywhere can we be more specific or are we sure that Staff and TA will only be there only once in a single span? because this test will pass even if it find staff and TA anywhere in the alert banner.

A-ASAD: There cannot be these words unless someone includes these in their name. Yes, the test will pass if there’s such a case. Updated it.

PR #86, opened 2022-03-08
fix: report indicator showing different on post summary

muhammadadeeltajamul:
https://openedx.atlassian.net/browse/TNL-9648

Before
tnl-9648-before

After
tnl-9648-after

### DIFF
--- a/src/discussions/posts/post/PostLink.jsx
+++ b/src/discussions/posts/post/PostLink.jsx
@@ -5,7 +5,7 @@ import { Link } from 'react-router-dom';
 
 import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 import { Badge, Icon } from '@edx/paragon';
-import { Flag, Pin } from '@edx/paragon/icons';
+import { Pin } from '@edx/paragon/icons';
 
 import { Routes, ThreadType } from '../../../data/constants';
 import AuthorLabel from '../../common/AuthorLabel';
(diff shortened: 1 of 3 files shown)


ayub02:
@muhammadadeeltajamul kindly review the font type, size and styling for "Reported". Seems a bit odd when compared to Figma mockups.

AhtishamShahid (approved):

awais-ansari (approved):

muhammadadeeltajamul:
> @muhammadadeeltajamul kindly review the font type, size and styling for "Reported". Seems a bit odd when compared to Figma mockups.

@ayub02 "Reported" text has all the same properties as of "Answered" text in mockups. I used all the properties of "Answered" in "Reported". If you think that we need to change this one or both, let me know. I will update it accordingly.

awais-ansari:
> > @muhammadadeeltajamul kindly review the font type, size and styling for "Reported". Seems a bit odd when compared to Figma mockups.
> 
> @ayub02 "Reported" text has all the same properties as of "Answered" text in mockups. I used all the properties of "Answered" in "Reported". If you think that we need to change this one or both, let me know. I will update it accordingly.

@ayub02 font size, font family and background color are different because of the theme. @muhammadadeeltajamul attach his local env screenshot here, which does not have theme configure. If all the properties are same as mockup figma. It will work fine on stage env.

PR #95, opened 2022-03-18
fix: send request to fetch discussion settings only for admins

A-ASAD:
## Ticket: [TNL-9731](https://openedx.atlassian.net/browse/TNL-9731)

Send a request to fetch discussion settings only for admins. UI was breaking if a request is sent for other roles.


asadazam93 (approved):
Looks good. This should only be merged after the backend PR has been deployed

PR #107, opened 2022-03-28
fix: navbar shadow

awais-ansari:
[TNL-9693](https://openedx.atlassian.net/browse/TNL-9693)
- Fix Navbar bottom shadow

Fix
Screenshot 2022-03-28 at 8 42 27 PM
Figma
Screenshot 2022-03-28 at 8 41 41 PM

### DIFF
--- a/src/discussions/discussions-home/DiscussionsHome.scss
+++ b/src/discussions/discussions-home/DiscussionsHome.scss
@@ -1,0 +1,3 @@
+.nav-shadow {
+  box-shadow: 0px 2px 4px rgb(0 0 0 / 15%), 0px 2px 8px rgb(0 0 0 / 15%);
+}
--- a/src/discussions/discussions-home/DiscussionsHome.scss
+++ b/src/discussions/discussions-home/DiscussionsHome.scss
@@ -1,3 +1,0 @@
-.nav-shadow {
-  box-shadow: 0px 2px 4px rgb(0 0 0 / 15%), 0px 2px 8px rgb(0 0 0 / 15%);
-}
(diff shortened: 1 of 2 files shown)


src/discussions/discussions-home/DiscussionsHome.scss
```diff
@@ -0,0 +1,3 @@
+.nav-shadow {
+  box-shadow: 0px 2px 4px rgb(0 0 0 / 15%), 0px 2px 8px rgb(0 0 0 / 15%);
+}
```
mehaknasir, line 3: should we not made that fix in paragon?

AhtishamShahid: .shadow |   | box-shadow: 0 0.125rem 0.25rem rgba(0, 0, 0, 0.3) !important;
-- | -- | --
.shadow-lg |   | box-shadow: 0 0.25rem 0.5rem rgba(0, 0, 0, 0.3) !important;
.shadow-none |   | box-shadow: none !important;
.shadow-sm |   | box-shadow: 0 0.0625rem 0.125rem rgba(0, 0, 0, 0.2) !important;

These classes are available in paragon , Please verify if any is related probably `.shadow-sm`

https://paragon-openedx.netlify.app/foundations/css-utilities/

awais-ansari: Discussion Figma designs are using old paragon styles which are affected because the latest paragon changes. The above classes have been changed in the latest paragon and do not match with our Figma design. That's why I created a custom CSS class to meet Figma's requirement.

mehaknasir: I feel, I may be wrong but we should go with the updated thing and ask Jonathan on this one what he suggests.

A-ASAD (approved):

src/discussions/discussions-home/DiscussionsHome.scss
awais-ansari: According to Jonathan, He is using shadow elevation level 2 in this design. Shadow elevation effects exist in paragon as SCSS mixin `pgn-box-shadow`. This mixin is available in only paragon but not available as a class to use across platforms. It will be implemented in paragon's future milestones. So, we decided to add a custom style here and replace it with paragon class after its paragon implementation.

PR #139, opened 2022-04-18
fix: Updated dependencies for Node 16

Jawayria:
Updated dependencies and regenerated package-lock file with NPM8


mehaknasir:
duplicate #140

PR #142, opened 2022-04-21
fix: hide edit reason dropdown while adding a comment or response

awais-ansari:
### [INF-149](https://openedx.atlassian.net/browse/INF-149)
- Hide the `Reason for editing` dropdown when the user (with moderator role and not a post/comment author) adds a response or a comment

**Add Response**
Screenshot 2022-04-21 at 4 29 30 PM

**Add Comment**
Screenshot 2022-04-21 at 10 04 51 PM

### DIFF
--- a/src/discussions/comments/comment/CommentEditor.jsx
+++ b/src/discussions/comments/comment/CommentEditor.jsx
@@ -20,6 +20,7 @@ function CommentEditor({
   intl,
   comment,
   onCloseEditor,
+  editing,
 }) {
   const { authenticatedUser } = useContext(AppContext);
   const userIsPrivileged = useSelector(selectUserIsPrivileged);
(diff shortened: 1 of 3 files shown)


src/discussions/comments/comment/CommentEditor.jsx
```diff
@@ -142,6 +145,11 @@ CommentEditor.propTypes = {
   }).isRequired,
   onCloseEditor: PropTypes.func.isRequired,
   intl: intlShape.isRequired,
+  edit: PropTypes.bool,
+};
+
+CommentEditor.defaultProps = {
+  edit: true,
```
mehaknasir, line 152: can we make default edit: false , and pass edit prop only when we need it to be true?

mehaknasir (commented):
just a suggestion related to react convention, otherwise looks good to me.

src/discussions/comments/comment/CommentEditor.jsx
awais-ansari: This component's primary purpose is editing and is always used for it. This component is also used in many components for editing propose. So default behavior should be editing true.  Only two use cases are not adding the edit reason.

mehaknasir: right.

mehaknasir (approved):

A-ASAD (approved):

PR #165, opened 2022-05-25
fix: image rendering fix for those uploaded using legacy experience

mehaknasir:
TICKET: https://2u-internal.atlassian.net/browse/INF-231

Image loaded from the legacy experience was not rendering in new MFE due to a change made for rendering MathML in a previous PR. as we no longer using MathML syntax and wiris support has been dropped that change is reverted and MFE will again use `renderedBody` key from API to preview post in different areas, and the ripple effect is reverted. now image uploaded from legacy will be visible in new discussion Experience.

### DIFF
--- a/src/discussions/posts/post/Post.jsx
+++ b/src/discussions/posts/post/Post.jsx
@@ -75,7 +75,7 @@ function Post({
       </div>
       <PostHeader post={post} actionHandlers={actionHandlers} />
       <div className="d-flex my-2 text-break">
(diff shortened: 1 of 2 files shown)


muhammadadeeltajamul (approved):

asadazam93 (approved):

asadazam93:
@mehaknasir Can you please fix the conflicts?

AhtishamShahid (approved):

PR #187, opened 2022-06-08
fix: tags added for reported/pinned/read/selected posts

mehaknasir:
Tickets: https://2u-internal.atlassian.net/browse/INF-278, https://2u-internal.atlassian.net/browse/INF-276

Accessibility issues are resolved in this PR.


awais-ansari (approved):

muhammadadeeltajamul (approved):

wittjeff (approved):

wittjeff:
I'm not 100% sure that the sr spans will read correctly here, but we can try it this way. Sometimes when you have a <span class='sr'>foo</>, the 'foo' merges with the previous word (usually visible) and may cause a spelling error when the string is sent to the TTS speech engine.  The solution is just to put a space before the first word within the sr span.

mehaknasir:
> I'm not 100% sure that the sr spans will read correctly here, but we can try it this way. Sometimes when you have a foo</>, the 'foo' merges with the previous word (usually visible) and may cause a spelling error when the string is sent to the TTS speech engine. The solution is just to put a space before the first word within the sr span.

I have handled it according to the provided feedback

PR #201, opened 2022-06-22
feat: clickable username in post summery card

awais-ansari:
### [INF-243](https://2u-internal.atlassian.net/browse/INF-243)
1. Click on post summery card username will redirect to the user all created and interacted posts under the learner tab.

### DIFF
--- a/src/discussions/learners/LearnerPostsView.jsx
+++ b/src/discussions/learners/LearnerPostsView.jsx
@@ -34,13 +34,14 @@ function LearnerPostsView({ intl }) {
   const { courseId, learnerUsername: username } = useContext(DiscussionContext);
   const nextPage = useSelector(selectThreadNextPage());
   const { id: userId } = useSelector(selectLearnerProfile(username));
+  const authorId = history.location.state?.authorId;
 
   useEffect(() => {
-    dispatch(fetchUserPosts(courseId, username, userId));
+    dispatch(fetchUserPosts(courseId, username, userId || authorId));
   }, [courseId, username]);
 
   const loadMorePosts = () => (
-    dispatch(fetchUserPosts(courseId, username, userId, {
+    dispatch(fetchUserPosts(courseId, username, userId || authorId, {
       page: nextPage,
     }))
   );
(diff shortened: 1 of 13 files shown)


src/discussions/learners/LearnerPostsView.jsx
```diff
@@ -33,13 +33,14 @@ function LearnerPostsView({ intl }) {
   const { courseId, learnerUsername: username } = useContext(DiscussionContext);
   const nextPage = useSelector(selectThreadNextPage());
   const { id: userId } = useSelector(selectLearnerProfile(username));
+  const authorId = history.location.state?.authorId;
 
   useEffect(() => {
-    dispatch(fetchUserPosts(courseId, username, userId));
+    dispatch(fetchUserPosts(courseId, username, userId || authorId));
```
asadazam93, line 39: is there a case where the first userId will be null?

src/discussions/learners/LearnerPostsView.jsx
```diff
@@ -88,7 +89,7 @@ function LearnerPostsView({ intl }) {
             <Spinner animation="border" variant="primary" size="lg" />
           </div>
         ) : (
-          nextPage && (
+          nextPage && loadingStatus === RequestStatus.SUCCESSFUL && (
```
asadazam93, line 92: is this change related? Or does this fix some other issue?

src/discussions/posts/post/PostLink.test.jsx
```diff
@@ -15,7 +15,7 @@ function renderComponent(post) {
   return render(
     <IntlProvider locale="en">
       <AppProvider store={store}>
-        <PostLink post={post} key={post.id} isSelected={() => true} />
+        <PostLink post={post} key={post.id} isSelected={() => true} learnerTab />
```
mehaknasir, line 18: all test cases here assuming component called from learner tab? whereas this prop is not required to be passed in the component. why do we need that change in test cases?

src/discussions/learners/LearnerPostsView.jsx
mehaknasir: this is an unrelated change as seen from PR description. I think it saves us from making extra call when loading new thread is already in progress.

mehaknasir (approved):
just a question posted, rest of the structure looks good to me. also if there is some fixes apart  from the actual ticket kindly add in PR description

asadazam93:
@awais-ansari the PR requires rebase

src/discussions/posts/post/PostLink.test.jsx
awais-ansari: Yes, you are right. I'll make it dynamic which will cover both cases leaner and simple posts.

asadazam93 (approved):

PR #203, opened 2022-06-23
fix: filtering posts in a topic doesn't work when filtered results are empty [BD-38] []

xitij2000:
A list of all posts in a topic is stored in an object for quick lookup. This is used when switching between topics to get the list of posts in that topic. When filters change, the list of posts in a topic is updated and this is picked up by the UI. If changing filters when browsing a topic results in an empty result, this would cause the original list of topics for that topic to be retained, causing the filter to simply not apply. This commit fixes that by resetting the post list in that case.

**Testing instructions**

Before checking out this PR try the following:
- In an empty topic, create a few discussion posts. 
- Navigate to that topic
- Filter that topic by "discussions"
- Now filter by "questions"
- The "questions" filter will not apply. 
- Now try creating one or more question posts in the same topic. 
- Switch between the "discussion" and "question" filter again. It should work now. 

Now try the above after checking out this PR. Switching between "discussions" and "questions" will work now, even if one of the filters is empty.

### DIFF
--- a/src/discussions/posts/data/api.js
+++ b/src/discussions/posts/data/api.js
@@ -27,6 +27,7 @@ export const coursesApiUrl = `${apiBaseUrl}/api/discussion/v1/courses/`;
  * @param {string} threadType Can be 'discussion' or 'question'.
  * @param {ThreadViewStatus} view Set to "unread" on "unanswered" to filter to only those statuses.
  * @param {boolean} countFlagged If true, abuseFlaggedCount will be available.
+ * @param {number} cohort
  * @returns {Promise<{}>}
  */
 export async function getThreads(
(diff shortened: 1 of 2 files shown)


openedx-webhooks:
Thanks for the pull request, @xitij2000!


asadazam93 (approved):

openedx-webhooks:
@xitij2000 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/frontend-app-discussions&entry.867055334=https://github.com/openedx/frontend-app-discussions/pull/203&entry.1484655318=https://github.com/xitij2000&entry.752974735=2022-06-23+09:16&entry.1917517419=2022-06-24+08:56&entry.2133058324=Yes) so we can improve your experience in the future.

awais-ansari (approved):

PR #208, opened 2022-07-05
fix: set default value for cohorts dropdown

awais-ansari:
### [INF-329](https://2u-internal.atlassian.net/browse/INF-329)

- Set `All learners` as the default value for cohorts dropdown 
- Cohorts dropdown is only visible in `add new post` for privileged user
- Cohort can not be changed while editing a post for all users

### Before fix
Screenshot 2022-07-05 at 5 34 20 PM

### After fix
Screenshot 2022-07-05 at 5 28 14 PM
Screenshot 2022-07-05 at 5 28 44 PM

### DIFF
--- a/src/discussions/posts/post-editor/PostEditor.jsx
+++ b/src/discussions/posts/post-editor/PostEditor.jsx
@@ -124,6 +124,8 @@ function PostEditor({
     }
     dispatch(hidePostEditor());
   };
+  // null stands for no cohort restriction ("All learners" option)
+  const selectedCohort = (cohort) => (cohort === 'default' ? null : cohort);
 
   const submitForm = async (values) => {
     if (editExisting) {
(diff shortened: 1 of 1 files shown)


src/discussions/posts/post-editor/PostEditor.jsx
```diff
@@ -281,24 +281,24 @@ function PostEditor({
...
+              <Form.Control
+                name="cohort"
+                as="select"
+                value={values.cohort}
+                onChange={handleChange}
+                onBlur={handleBlur}
+                aria-describedby="cohortAreaInput"
+                floatingLabel={intl.formatMessage(messages.cohortVisibility)}
+              >
+                <option value="default">{intl.formatMessage(messages.cohortVisibilityAllLearners)}</option>
```
asadazam93, line 295: In case of all learners selected the value sent to the backend will be "default"

awais-ansari: No, a Null value will be sent in the case of all learners selected.

mehaknasir (approved):

PR #212, opened 2022-07-19
fix: updated pinned post icon

muhammadadeeltajamul:
Ticket Link:
https://2u-internal.atlassian.net/browse/INF-354

new_pinned_icon

### DIFF
--- a/src/discussions/posts/post/PostLink.jsx
+++ b/src/discussions/posts/post/PostLink.jsx
@@ -6,7 +6,7 @@ import { Link } from 'react-router-dom';
 
 import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 import { Badge, Icon } from '@edx/paragon';
-import { Bookmark } from '@edx/paragon/icons';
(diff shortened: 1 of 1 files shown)


src/discussions/posts/post/PostLink.jsx
```diff
@@ -95,6 +90,10 @@ function PostLink({
                   <span className="sr-only">{' '}reported</span>
                 </Badge>
               )}
+
+              {post.pinned && (
+                <Icon src={PushPin} className={postReported || showAnsweredBadge ? 'ml-2' : 'ml-auto'} />
+              )}
```
awais-ansari, line 96: We are using 20x20 icons in the post summary card. Please use the `icon-size` class which will set the icon height and width. Please make sure the height should not exceed 22px.

awais-ansari (changes requested):

awais-ansari (approved):

mehaknasir (approved):

PR #213, opened 2022-07-19
fix: accessibility fix added for the threads section

mehaknasir:
Ticket ID: https://2u-internal.atlassian.net/browse/INF-333

DOM PRINT AFTER FIX
- heading moved up 
- no interstitial divs between list items
- list / list items roles added for nested list items
- sr only description added for replies to response comments.

Screenshot 2022-07-19 at 11 44 06 PM

### DIFF
--- a/src/discussions/comments/CommentsView.jsx
+++ b/src/discussions/comments/CommentsView.jsx
@@ -66,34 +66,37 @@ function DiscussionCommentsView({
     handleLoadMoreResponses,
   } = usePostComments(postId, endorsed);
   return (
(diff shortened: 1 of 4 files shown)


awais-ansari (approved):

asadazam93 (approved):

wittjeff:
<li> is first child of the <ul>

This one part does not seem to be satisfied in the PR.  Could you please move the four divs that are above the first div role=”listitem” above the div role=”list”?Otherwise it looks just right.  Thanks!

PR #215, opened 2022-07-22
fix: load thread if thread data doesnot exist in redux

muhammadadeeltajamul:
Thread is not loaded when thread data is not found in the redux store. After this change, it will send an API call to get thread from server.

This issue was found when working on the following ticket
https://2u-internal.atlassian.net/browse/INF-349


awais-ansari (approved):
Please test is change properly with different users and use cases.

mehaknasir (approved):

asadazam93 (approved):

PR #223, opened 2022-07-28
 [Blocked] Added count_flagged param to learners tab

AhtishamShahid:
Show reported tag on post link if comment on the post is flagged
The learner tab was using `fetchUserPosts` API it  does not support `count_flagged` now It is updated to use `fetchThreads`.

https://2u-internal.atlassian.net/browse/INF-365


awais-ansari (commented):
@AhtishamShahid I'm not sure that FetchThreads API will return posts that are interacted with by a user.
According to my understanding, FetchThreads API returns posts that are related/created by a user.

PR #237, opened 2022-08-05
fix: last login time is hidden from FE

mehaknasir:
https://2u-internal.atlassian.net/browse/INF-427

- last login time is hidden from FE and will be added back once we have an actual time of last activity on forum starts returning.


awais-ansari (approved):

AhtishamShahid (approved):

awais-ansari (commented):
@mehaknasir can you please look into fail test cases

PR #238, opened 2022-08-05
fix: add a response button added at the end of thread

mehaknasir:
https://2u-internal.atlassian.net/browse/INF-407

Screenshot 2022-08-06 at 11 56 22 PM

> @mehaknasir Button's top/ bottom margin is not matching with Figma. Please match it from Figma.

fixed it ss attached

### DIFF
--- a/src/discussions/comments/CommentsView.jsx
+++ b/src/discussions/comments/CommentsView.jsx
@@ -76,7 +76,8 @@ function DiscussionCommentsView({
         {comments.map(comment => (
           <Comment comment={comment} key={comment.id} postType={postType} />
         ))}
-
+        {!!comments.length
+          && <ResponseEditor postId={postId} addWrappingDiv />}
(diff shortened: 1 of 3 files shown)


awais-ansari (approved):
@mehaknasir Button's top/ bottom margin is not matching with Figma. Please match it from Figma.

src/discussions/comments/CommentsView.jsx
```diff
@@ -76,7 +76,8 @@ function DiscussionCommentsView({
         {comments.map(comment => (
           <Comment comment={comment} key={comment.id} postType={postType} />
         ))}
-
+        {!!comments.length
```
asadazam93, line 79: Why the trailing !! ?

mehaknasir: because it adds length count in dom. it is to change it to boolean

PR #239, opened 2022-08-05
fix: added no results bar to topics tab and search info to learners tab

muhammadadeeltajamul:
- Added NoResults to Topics and Learners tab
- Added SearchInfo to Learners tab
- Hide filter bar when user searches in Learners tab

Ticket Link:
https://2u-internal.atlassian.net/browse/INF-412
https://2u-internal.atlassian.net/browse/INF-414

### DIFF
--- a/src/components/Search.jsx
+++ b/src/components/Search.jsx
@@ -1,4 +1,4 @@
-import React, { useContext, useEffect } from 'react';
+import React, { useContext, useEffect, useMemo } from 'react';
 
 import camelCase from 'lodash/camelCase';
 import { useDispatch, useSelector } from 'react-redux';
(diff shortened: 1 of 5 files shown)


mehaknasir (approved):

src/components/Search.jsx
```diff
@@ -18,9 +18,20 @@ function Search({ intl }) {
...
 
+  let currentValue = '';
+  if (isPostSearch) {
+    currentValue = postSearch;
+  } else if (isTopicSearch) {
+    currentValue = topicSearch;
+  } else {
+    currentValue = learnerSearch;
+  }
+
```
awais-ansari, line 34: convert into memoize function. Otherwise, this calculation will happen on every render.

const currentValue = useMemo(()=> {
let value = '';
  if (isPostSearch) {
    value = postSearch;
  } else if (isTopicSearch) {
    value = topicSearch;
  } else {
    value = learnerSearch;
  }
return value;
}, [isPostSearch, isTopicSearch])

awais-ansari (approved):

PR #259, opened 2022-08-19
fix: no result found text remove when call in progress

mehaknasir:
No result found text will not be visible in case an API call is in progress.

https://user-images.githubusercontent.com/67791278/185651852-4fedf6e1-e979-48c2-96f7-a301dfa3a58a.mov


AhtishamShahid (approved):

AhtishamShahid (commented):
Please add a test case for this as well. other than that Looks good,

mehaknasir:
> Please add a test case for this as well. other than that Looks good,

the test cases for the whole component do not exist, this will be covered in the ticket we have for test coverage improvements exercise

awais-ansari (approved):

PR #276, opened 2022-08-31
fix: learner tab is hidden from learners

mehaknasir:
- learner tab will not be visible to student roles 
- learners won't be able to access it through URL as well

### DIFF
--- a/src/discussions/common/AuthorLabel.jsx
+++ b/src/discussions/common/AuthorLabel.jsx
@@ -2,7 +2,6 @@ import React, { useContext } from 'react';
 import PropTypes from 'prop-types';
 
 import classNames from 'classnames';
-import { useSelector } from 'react-redux';
 import { Link, useLocation } from 'react-router-dom';
 
 import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
@@ -10,7 +9,7 @@ import { Icon } from '@edx/paragon';
 import { Institution, School } from '@edx/paragon/icons';
 
 import { Routes } from '../../data/constants';
-import { selectLearnersTabEnabled } from '../data/selectors';
+import { useShowLearnersTab } from '../data/hooks';
 import messages from '../messages';
 import { discussionsPath } from '../utils';
 import { DiscussionContext } from './context';
@@ -24,7 +23,6 @@ function AuthorLabel({
 }) {
   const location = useLocation();
   const { courseId } = useContext(DiscussionContext);
-  const learnersTabEnabled = useSelector(selectLearnersTabEnabled);
   let icon = null;
   let authorLabelMessage = null;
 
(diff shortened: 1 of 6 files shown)


src/discussions/data/hooks.js
```diff
@@ -167,3 +167,13 @@ export const useAlertBannerVisible = (content) => {
     || (content.abuseFlagged && canSeeReportedBanner)
   );
 };
+
+export const useIsShowLearnersTab = () => {
```
asadazam93, line 171: nit: useShowLearnersTab

src/discussions/data/hooks.js
```diff
@@ -167,3 +167,13 @@ export const useAlertBannerVisible = (content) => {
...
+  const learnersTabEnabled = useSelector(selectLearnersTabEnabled);
+  const userRoles = useSelector(selectUserRoles);
+  const isAdmin = useSelector(selectUserIsStaff);
+  const isGTA = useSelector(selectUserIsGroupTa);
```
asadazam93, line 175: IsGroupTA

src/discussions/data/hooks.js
```diff
@@ -167,3 +167,13 @@ export const useAlertBannerVisible = (content) => {
...
+  const privileged = useSelector(selectUserHasModerationPrivileges);
+  const allowedUsers = isAdmin || isGTA || privileged || (userRoles.includes('Student') && userRoles.length > 1);
```
asadazam93, line 177: Do we need the the first 3 checks here? If the userRoles includes Student and there is more than 1 role then the additional role will cover the first 3 checks here?

asadazam93 (approved):

awais-ansari (changes requested):
we also need to add a condition on the post summary card username redirection. Click on a post summary card username redirects a user to the learner area. So we need to disable the redirection link for learners and should show a plan text username.

src/discussions/common/AuthorLabel.jsx
```diff
@@ -41,6 +42,10 @@ function AuthorLabel({
 
   const className = classNames('d-flex align-items-center', labelColor);
 
+  const showUserNameAsLink = useShowLearnersTab()
+      && linkToProfile && author
+      && learnersTabEnabled && author !== messages.anonymous;
+
```
awais-ansari, line 48: You already checking the `learnersTabEnabled` in `useShowLearnersTab()`. So you can remove it here. 
nit: 
`const showUserNameAsLink = useShowLearnersTab() && linkToProfile && author 
&& author !== messages.anonymous;`

awais-ansari (approved):

src/discussions/common/AuthorLabel.jsx
mehaknasir: done

PR #297, opened 2022-09-21
Revert "fix: add response duplication button fixed"

asadazam93:
Reverts openedx/frontend-app-discussions#283


nedbat:
It would be great to include the reason for needing to revert.  It helps with understanding why the code is changing, and what is causing friction during development.

PR #307, opened 2022-09-30
fix: removed responses sort from frontend

muhammadadeeltajamul:
Removed responses sort from frontend

Ticket Link:
https://2u-internal.atlassian.net/browse/INF-546

### DIFF
--- a/src/discussions/comments/CommentsView.jsx
+++ b/src/discussions/comments/CommentsView.jsx
@@ -1,7 +1,6 @@
 import React, { useEffect, useMemo } from 'react';
 import PropTypes from 'prop-types';
 
-import { orderBy } from 'lodash';
 import { useDispatch, useSelector } from 'react-redux';
 import { useParams } from 'react-router';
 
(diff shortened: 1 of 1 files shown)


src/discussions/comments/CommentsView.jsx
```diff
@@ -67,8 +67,8 @@ function DiscussionCommentsView({
     isLoading,
     handleLoadMoreResponses,
   } = usePostComments(postId, endorsed);
-  const sortedComments = useMemo(() => orderBy(comments, ['endorsed', 'createdAt'],
-    ['desc', 'desc']));
+  const sortedComments = useMemo(() => orderBy(comments, ['endorsed'],
```
mehaknasir, line 70: Is sort based on endorsed working? What is the difference in ui behavior if we remove that as well?


Is it to get endorsed responses on top on initial load?
I think we are also working on shifting that logic to backend in another ticket.

mehaknasir (approved):

src/discussions/comments/CommentsView.jsx
muhammadadeeltajamul: Yes... Endorsed responses will be on top...
Frontend will be changed when the change will be implemented on backend

awais-ansari: We do not need to sortBy desc order. we can split the comments array into 2 arrays. one for endorse and one for un endorse and render these arrays one by one which will persist the API response order.  A filter function is available in utils file.

awais-ansari (approved):

PR #311, opened 2022-10-03
fix: remove editor text persistence

awais-ansari:
### [INF-589](https://2u-internal.atlassian.net/browse/INF-589)

- Text Editor will reset to the initial value then click on submit and cancel button while creating the post.
- Text Editor will reset to the initial value then click on submit and cancel button while creating the Response/Comment.
- Text Editor will not persist the text.

https://www.loom.com/share/896bd04092c1422aa864a8cd64e6256e

### DIFF
--- a/src/components/TinyMCEEditor.jsx
+++ b/src/components/TinyMCEEditor.jsx
@@ -92,7 +92,7 @@ export default function TinyMCEEditor(props) {
         browser_spellcheck: true,
         a11y_advanced_options: true,
         autosave_interval: '1s',
-        autosave_restore_when_empty: true,
+        autosave_restore_when_empty: false,
         plugins: 'autosave codesample link lists image imagetools code emoticons charmap',
         toolbar: 'undo redo'
           + ' | formatselect | bold italic underline'
--- a/src/discussions/posts/post-editor/PostEditor.jsx
+++ b/src/discussions/posts/post-editor/PostEditor.jsx
@@ -125,7 +125,21 @@ function PostEditor({
     const isCohorting = settings.alwaysDivideInlineDiscussions || settings.dividedInlineDiscussions.includes(tId);
     return isCohorting;
   };
-  const hideEditor = () => {
+
+  const initialValues = {
+    postType: post?.type || 'discussion',
+    topic: post?.topicId || topicId || nonCoursewareTopics?.[0]?.id,
+    title: post?.title || '',
+    comment: post?.rawBody || '',
+    follow: isEmpty(post?.following) ? true : post?.following,
+    anonymous: allowAnonymous ? false : undefined,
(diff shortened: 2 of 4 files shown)


muhammadadeeltajamul (approved):

src/index.scss
```diff
@@ -227,7 +227,7 @@ header {
 
 .header-action-bar {
   background-color: #fff;
-  z-index: 1;
+  z-index: 2;
```
saadyousafarbi, line 230: why is this changed?

src/discussions/posts/post-editor/PostEditor.jsx
```diff
@@ -188,18 +201,6 @@ function PostEditor({
...
-    topic: post?.topicId || topicId || nonCoursewareTopics?.[0]?.id,
-    title: post?.title || '',
-    comment: post?.rawBody || '',
-    follow: isEmpty(post?.following) ? true : post?.following,
-    anonymous: allowAnonymous ? false : undefined,
-    anonymousToPeers: allowAnonymousToPeers ? false : undefined,
-    editReasonCode: post?.lastEdit?.reasonCode || '',
-    cohort: post?.cohort || 'default',
-  };
-
```
mehaknasir, line 202: what happens? when you open the editor to edit posts/comments/responses?

src/components/TinyMCEEditor.jsx
```diff
@@ -92,7 +92,7 @@ export default function TinyMCEEditor(props) {
         browser_spellcheck: true,
         a11y_advanced_options: true,
         autosave_interval: '1s',
-        autosave_restore_when_empty: true,
+        autosave_restore_when_empty: false,
```
mehaknasir, line 95: why mark it as false when you can reset the form on cancel/submit?

mehaknasir (commented):
I feel like these changes will create more issues, like what if someone wrote the post of 100 lines and cancel it by mistake?  and similar usecases.

src/discussions/posts/post-editor/PostEditor.jsx
awais-ansari: It will work as it was working before. I just relocate this constant.

src/components/TinyMCEEditor.jsx
awais-ansari: Yes, I'm resetting the form state on cancel/submit but This plugin stores the draft value in browser storage. If it is true it will get the draft value from the browser storage after the editor initialization.

src/index.scss
awais-ansari: The text editor action bar and header action bar had the same z-index value `1` which can create unpredictable UI overlap glitches for some browsers that why I increase this value according to UI stack.

mehaknasir (approved):

PR #316, opened 2022-10-05
fix: added reportgmailerrored icon

muhammadadeeltajamul:
Need to add ReportGmailerrored icon to repo. It is not available in the paragon version currently being used in MFE

Ticket Link:
https://2u-internal.atlassian.net/browse/INF-554


muhammadadeeltajamul:
This fix is not needed because paragon version of MFE was upgraded to latest version.

PR #321, opened 2022-10-10
fix: fix for reported alert banner and reported badge

saadyousafarbi:
### Description

Update **Reported Alert Banner** and **Reported Badge** to use `abuseFlagged` param to show/hide Reported Label on content. 

The `abuseFlagged` param indicates the following for different roles:

- **Moderator Role:** **_Return True if content is reported by anyone._**
- **No Moderator Role:** _**Return True ONLY if content is reported by the user themselves.**_

We can avoid checking for different roles on the current user, and use the already calculated `abuseFlagged` value from backend.

### DIFF
--- a/src/discussions/common/AlertBanner.jsx
+++ b/src/discussions/common/AlertBanner.jsx
@@ -23,12 +23,11 @@ function AlertBanner({
   const { reasonCodesEnabled } = useSelector(selectModerationSettings);
   const userIsContentAuthor = getAuthenticatedUser().username === content.author;
   const canSeeLastEditOrClosedAlert = (userHasModerationPrivileges || userIsContentAuthor || userIsGroupTa);
-  const isReportedByCurrentUser = getAuthenticatedUser().username === content?.abuseFlaggedBy;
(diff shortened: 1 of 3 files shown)


muhammadadeeltajamul (approved):

src/discussions/common/AlertBanner.jsx
```diff
@@ -23,12 +23,11 @@ function AlertBanner({
   const { reasonCodesEnabled } = useSelector(selectModerationSettings);
   const userIsContentAuthor = getAuthenticatedUser().username === content.author;
   const canSeeLastEditOrClosedAlert = (userHasModerationPrivileges || userIsContentAuthor || userIsGroupTa);
-  const isReportedByCurrentUser = getAuthenticatedUser().username === content?.abuseFlaggedBy;
-  const canSeeReportedBanner = (userHasModerationPrivileges || userIsGroupTa || isReportedByCurrentUser);
+  const canSeeReportedBanner = content?.abuseFlagged;
```
mehaknasir, line 26: Can we rename it to canSeeReportedBadge Everywhere it is used for uniformity?

awais-ansari (approved):

mehaknasir (approved):

PR #324, opened 2022-10-11
chore: move feedback banner behind the DISPLAY_FEEDBACK_BANNER env variable

awais-ansari:
### [INF-590](https://2u-internal.atlassian.net/browse/INF-590)

- Move the feedback banner behind the DISPLAY_FEEDBACK_BANNER env variable
- Default `DISPLAY_FEEDBACK_BANNER` is false
- Banner will be visible when the `DISPLAY_FEEDBACK_BANNER` value will be `true`

### Dependency
https://github.com/edx/edge-internal/pull/471
https://github.com/edx/edx-internal/pull/7219


src/index.jsx
```diff
@@ -40,4 +41,12 @@ initialize({
...
   ],
+  handlers: {
+    config: () => {
+      mergeConfig({
+        LEARNING_BASE_URL: process.env.LEARNING_BASE_URL,
+        DISPLAY_FEEDBACK_BANNER: process.env.DISPLAY_FEEDBACK_BANNER || 'false',
+      }, 'DiscussionsConfig');
+    },
+  },
 });
```
mehaknasir, line 52: Why do we need this? I am already using env variables from edx-internal and I don't find the use of **`mergeconfig`** so I just skipped it before.

mehaknasir (approved):

src/index.jsx
mehaknasir: https://github.com/openedx/frontend-platform#application-configuration

 i got its usage.

saadyousafarbi (approved):

PR #331, opened 2022-10-18
fix: convert clickable author label to text in posts summary list

awais-ansari:
### [INF-596](https://2u-internal.atlassian.net/browse/INF-596)

- User name is no more clickable in the posts summary list in all 4 tabs
- Add new test cases for an author label component 

https://www.loom.com/share/e191efd27a744ac180d57cb0109de955

### DIFF
--- a/src/discussions/common/AuthorLabel.jsx
+++ b/src/discussions/common/AuthorLabel.jsx
@@ -40,7 +40,7 @@ function AuthorLabel({
   const className = classNames('d-flex align-items-center', labelColor);
 
   const showUserNameAsLink = useShowLearnersTab()
-      && linkToProfile && author && author !== messages.anonymous;
+      && linkToProfile && author && author !== 'anonymous';
 
   const labelContents = (
     <div className={className}>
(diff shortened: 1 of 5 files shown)


src/discussions/common/AuthorLabel.jsx
```diff
@@ -40,7 +40,7 @@ function AuthorLabel({
   const className = classNames('d-flex align-items-center', labelColor);
 
   const showUserNameAsLink = useShowLearnersTab()
-      && linkToProfile && author && author !== messages.anonymous;
+      && linkToProfile && author && author !== 'anonymous';
```
mehaknasir, line 43: are you sure the author name is returned as anonymous in data?

mehaknasir (approved):

src/discussions/common/AuthorLabel.jsx
awais-ansari: Yes, It will be anonymous

ayeshoali (approved):

PR #363, opened 2022-11-25
docs: Include in Open edX release

arbrandes:
Commit to including the MFE in Open edX releases going forward.


arbrandes:
Since technically this is a minor change with no production impact and we're running out of time to include this MFE in Olive, i'm going to go ahead and merge this.

nedbat:
fwiw, "docs:" should be changes to words that people read.  This is "build:".

PR #367, opened 2022-11-30
fix: removed side bar incase there are no posts

AhtishamShahid:
### **[INF-625](https://2u-internal.atlassian.net/browse/INF-625)**
- Removed the Sidebar in case there are no posts.

Screen Shot 2022-11-30 at 12 23 57 PM

https://user-images.githubusercontent.com/26253150/207581845-2a976e50-e2ff-4276-9fd0-aca056f3716a.mov

### DIFF
--- a/src/discussions/data/hooks.js
+++ b/src/discussions/data/hooks.js
@@ -25,13 +25,12 @@ import {
   selectIsCourseStaff,
   selectLearnersTabEnabled,
   selectModerationSettings,
-  selectPostThreadCount, selectThreadCallStatus,
+  selectPostThreadCount,
   selectUserHasModerationPrivileges,
   selectUserIsGroupTa,
   selectUserIsStaff,
 } from './selectors';
 import { fetchCourseConfig } from './thunks';
-import { selectLearnerPages } from '../learners/data/selectors';
 
 export function useTotalTopicThreadCount() {
   const topics = useSelector(selectTopics);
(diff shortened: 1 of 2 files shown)


src/discussions/data/hooks.js
```diff
@@ -50,7 +50,7 @@ export const useSidebarVisible = () => {
     return true;
   }
 
-  if (isViewingTopics || isViewingLearners) {
+  if ((isViewingTopics || isViewingLearners) && totalThreads > 0) {
```
awais-ansari, line 53: @AhtishamShahid This solution is not an optimal solution. This change is breaking things in locations. I checkout your branch on my local machine and this is happing. Please check the video. 
https://www.loom.com/share/27abf0da19424bd38c8daef5673b4a39

awais-ansari (changes requested):

src/discussions/data/hooks.js
```diff
@@ -21,11 +21,12 @@ import { discussionsPath } from '../utils';
 import {
   selectAreThreadsFiltered, selectLearnersTabEnabled,
   selectModerationSettings,
-  selectPostThreadCount,
+  selectPostThreadCount, selectThreadCallStatus,
```
awais-ansari, line 24: Test cases are failing and the selector is not found in the relative file. Please attach the before, and after videos.

awais-ansari (changes requested):

ayeshoali (approved):

PR #377, opened 2022-12-07
feat: Support runtime configuration

arbrandes:
frontend-platform supports runtime configuration since 2.5.0 (see [the PR](https://github.com/openedx/frontend-platform/pull/335) that introduced it), but it requires MFE cooperation.  This implements just that: by avoiding making configuration values constant, it should now be possible to change them after initialization.

Almost all changes here relate to the `LMS_BASE_URL` setting, which in most places was treated as a constant.

### Testing

This should be a NOOP for any deployments of the MFE that _don't_ make explicit use of runtime configuration.  This is reflected in all unit tests, which needed little to no significant modifications.  This change should have no effect on usage of this MFE in the devstack or in current production deployments.

To actually see the desired effects of the change, one needs to deploy Tutor with the [correspondingly modified tutor-mfe plugin](https://github.com/overhangio/tutor-mfe/pull/91).  Where before several features would fail (adding a post, navigating to the topics page, etc), they should all now work.

### DIFF
--- a/src/components/NavigationBar/data/api.js
+++ b/src/components/NavigationBar/data/api.js
@@ -2,7 +2,7 @@
 import { camelCaseObject } from '@edx/frontend-platform';
 import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
 
-import { API_BASE_URL } from '../../../data/constants';
+import { getApiBaseUrl } from '../../../data/constants';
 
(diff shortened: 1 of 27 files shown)


awais-ansari (approved):

arbrandes:
Thanks for the review, @awais-ansari!  Much appreciated!

PR #378, opened 2022-12-08
feat: added archived topics is post editor

AhtishamShahid:
## Ticket 
https://2u-internal.atlassian.net/browse/INF-672

## Description 

Added archived topics in the post editor for non-learner users and archived topics would only be visible only if  in-context is enabled and `inContext` parameter is present in the app URL.

### DIFF
--- a/src/discussions/posts/post-editor/PostEditor.jsx
+++ b/src/discussions/posts/post-editor/PostEditor.jsx
@@ -21,6 +21,7 @@ import { TinyMCEEditor } from '../../../components';
 import FormikErrorFeedback from '../../../components/FormikErrorFeedback';
 import PostPreviewPane from '../../../components/PostPreviewPane';
 import { useDispatchWithState } from '../../../data/hooks';
+import { selectArchivedTopics } from '../../../data/selectors';
 import { selectCourseCohorts } from '../../cohorts/data/selectors';
 import { fetchCourseCohorts } from '../../cohorts/data/thunks';
 import { DiscussionContext } from '../../common/context';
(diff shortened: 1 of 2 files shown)


awais-ansari (changes requested):
@AhtishamShahid this achieved topics selector is related to v2Topics API which will deprecate after v3Topics API structure implementation. I think you need to implement achieved topics in V3 API.

AhtishamShahid:
> 

What is the time frame of depreciation?

awais-ansari:
> > 
> 
> What is the time frame of depreciation? and Implementing v3 API would be out of the scope of this ticket.

@AhtishamShahid V2 will deprecate in this sprint. I'm working on the V3 topic structure implementation right now after that v2 structure will not be able further.

AhtishamShahid:
> > > 
> > 
> > 
> > What is the time frame of depreciation? and Implementing v3 API would be out of the scope of this ticket.
> 
> @AhtishamShahid V2 will deprecate in this sprint. I'm working on the V3 topic structure implementation right now after that v2 structure will not be able further.

I guess we should block this ticket because it is using the same redux state where currently topics are stored until your work on v3 is done.

awais-ansari:
> > > > 
> > > 
> > > 
> > > What is the time frame of depreciation? and Implementing v3 API would be out of the scope of this ticket.
> > 
> > 
> > @AhtishamShahid V2 will deprecate in this sprint. I'm working on the V3 topic structure implementation right now after that v2 structure will not be able further.
> 
> I guess we should block this ticket because it is using the same redux state where currently topics are stored until your work on v3 is done.

@AhtishamShahid Yes, For now, you can block the frontend ticket but you can confirm the logic of the archived topic from the backend. If the backend v3 topics API already has the logic of archived topics then you will only be blocked on the frontend side till next week.

PR #380, opened 2022-12-08
fix: image paste is blocked in tinymce

mehaknasir:
### [inf-670](https://2u-internal.atlassian.net/browse/INF-670)

- the issue was appearing on pasting images the upload image function was not being called but the image looks to be pasted properly, we have disabled the paste image option to avoid confusion about the image being added in the post which actually was not uploaded


src/components/TinyMCEEditor.jsx
```diff
@@ -100,12 +101,13 @@ export default function TinyMCEEditor(props) {
           skin: false,
           menubar: false,
           branding: false,
+          paste_data_images: false,
```
asadazam93, line 104: When the user pastes an image nothing will be pasted?

asadazam93 (approved):

awais-ansari (approved):

PR #386, opened 2022-12-15
feat: Support runtime configuration (second attempt)

arbrandes:
(This reintroduces the change in 9f84230c that was later reverted by 67b0b33a.)

frontend-platform supports runtime configuration since 2.5.0 (see the PR that introduced it[1], but it requires MFE cooperation.  This implements just that: by avoiding making configuration values constant, it should now be possible to change them after initialization.

Almost all changes here relate to the `LMS_BASE_URL` setting, which in most places was treated as a constant.

[1] https://github.com/openedx/frontend-platform/pull/335

### DIFF
--- a/src/components/NavigationBar/data/api.js
+++ b/src/components/NavigationBar/data/api.js
@@ -2,7 +2,7 @@
 import { camelCaseObject } from '@edx/frontend-platform';
 import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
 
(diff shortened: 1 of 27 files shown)


asadazam93:
@arbrandes have you updated anything in this PR? OR do you want to just merge and try to debug on stage?

awais-ansari (approved):

arbrandes:
As noted on the parent issue, no changes here yet.  Going to do another visual pass.

PR #389, opened 2022-12-20
fix: loading icon and filter menu shows when api call is in progress

ayeshoali:
[INF-683](https://2u-internal.atlassian.net/browse/INF-683)

**Description**
Loading icon and filter menu shows when api call is in progress in My posts tab.

Before

My posts tab

After
chrome-capture-2022-11-20

### DIFF
--- a/src/discussions/posts/data/slices.js
+++ b/src/discussions/posts/data/slices.js
@@ -57,7 +57,6 @@ const threadsSlice = createSlice({
       if (state.author !== payload.author) {
         state.pages = [];
         state.author = payload.author;
-        state.totalThreads = null;
       }
       state.status = RequestStatus.IN_PROGRESS;
     },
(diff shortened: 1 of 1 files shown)


src/discussions/posts/data/slices.js
```diff
@@ -57,7 +57,6 @@ const threadsSlice = createSlice({
       if (state.author !== payload.author) {
         state.pages = [];
         state.author = payload.author;
-        state.totalThreads = null;
```
mehaknasir, line 60: I feel like this fix will have ripple issues as  this is emptying older threads count when a new request is made and if at certain point the call will fail, redux will still be populated with outdated data

mehaknasir (approved):

muhammadadeeltajamul (approved):

awais-ansari (approved):

PR #393, opened 2022-12-23
style: z-index fixes for UI elements in user dropdown and editor toolbar

mehaknasir:
After Fix:

The toolbar is not going over the sticky top also the user menu drop-down stays above the sticky header now.

https://user-images.githubusercontent.com/67791278/209301165-1bce0518-aabb-43ee-9966-1d1b65831694.mov

### DIFF
--- a/src/index.scss
+++ b/src/index.scss
@@ -117,9 +117,6 @@ $fa-font-path: "~font-awesome/fonts";
 }
 
 header {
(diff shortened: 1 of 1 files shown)


asadazam93 (approved):

awais-ansari (approved):
@mehaknasir Please make sure the file has the new line at the end.

PR #419, opened 2023-01-26
fix: use threads api in case of search

AhtishamShahid:
## Description 
On my posts page search was using `fetchUserPosts` API which does not allow search by text so changed code to use `fetchThreads` API so it would filter by search text and author . 

## Ticket 

https://2u-internal.atlassian.net/browse/INF-734

### DIFF
--- a/src/discussions/posts/PostsList.jsx
+++ b/src/discussions/posts/PostsList.jsx
@@ -1,5 +1,5 @@
 import React, {
-  useCallback, useContext, useEffect, useMemo,
+  useCallback, useContext, useEffect, useMemo, useRef,
 } from 'react';
 import PropTypes from 'prop-types';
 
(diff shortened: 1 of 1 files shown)


mehaknasir (approved):

src/discussions/posts/PostsList.jsx
```diff
@@ -51,10 +52,11 @@ function PostsList({
       isFilterChanged,
     };
 
+    if ((prevFilters.current.search !== '' && prevFilters.current.search !== filters.search) || !showOwnPosts) {
+      dispatch(fetchThreads(courseId, params));
+    }
```
awais-ansari, line 57: This part of the condition `(prevFilters.current.search !== '' && prevFilters.current.search !== filters.search)` would be false always. `prevFilters` is initialized with an empty string and does not change after that. So its value remains an empty string and this check `prevFilters.current.search !== ''` will get false. 
Indirectly,  Your check only depends on `!showOwnPosts`. The previous code was working on the same logic. However, current change is not required.

awais-ansari (changes requested):

awais-ansari (approved):

PR #421, opened 2023-01-26
Revert "feat: added mathjax v3 support in platform"

asadazam93:
Reverts openedx/frontend-app-discussions#420

Reverting this PR as it broke the UI


mehaknasir (approved):

mehaknasir:
will recreate PR after fixing the issue of the broken UI. revert is necessary as it is breaking the comments/responses feature

PR #422, opened 2023-01-26
Mathax version update

mehaknasir:
- this adds back mathjax support in platform
- this fixes response/comments overwrite  issue


mehaknasir:
#423 duplicate as branch is outdated

PR #457, opened 2023-02-28
feat: send enableInContextSidebar param to backend to identify source of content for events

saadyousafarbi:
### Description

send `enableInContextSidebar` param to backend to identify source of content for events. We are doing this to provide flexibility in terms of events analysis (to determine generation of content through new in-context sidebar)

Targeted events are as follow:

- `edx.forum.thread.created`
- `edx.forum.response.created`
- `edx.forum.comment.created`
- `edx.forum.thread.viewed`

Related PR on platform -- https://github.com/openedx/edx-platform/pull/31849

### DIFF
--- a/src/discussions/posts/data/thunks.js
+++ b/src/discussions/posts/data/thunks.js
@@ -204,6 +204,7 @@ export function createNewThread({
   anonymous,
   anonymousToPeers,
   cohort,
+  enableInContextSidebar,
 }) {
   return async (dispatch) => {
     try {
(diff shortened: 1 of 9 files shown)


ayeshoali (approved):

src/discussions/posts/data/thunks.js
```diff
@@ -223,7 +224,7 @@ export function createNewThread({
         following,
         anonymous,
         anonymousToPeers,
-      });
+      }, enableInContextSidebar);
```
AhtishamShahid, line 227: DiscussionContext can be directly imported to use it here, it would be a cleaner approach, and will not require changes in additional files. i.e `src/discussions/post-comments/comments/comment/CommentEditor.jsx`


```js
import { useContext } from 'react';
import { DiscussionContext } from '../../common/context';


const { enableInContextSidebar } = useContext(DiscussionContext);

awais-ansari: @AhtishamShahid hooks can only be used in the components. Unfortunately, the above approach won't work here.

awais-ansari (approved):

PR #482, opened 2023-03-28
fix: typeset failed: window.MathJax.typesetClear is not a function

ayeshoali:
[INF-840](https://2u-internal.atlassian.net/browse/INF-840)
Description

Currently, Discussions MFE has many frequent JS errors on the new relic dashboard.
This PR intends to resolve `typeset failed:e.typesetPromise is not a function`

#### Post-merge Checklist

* [ ] Deploy the changes to prod after verifying on stage or ask **@openedx/edx-infinity** to do it. 
* [ ] 🎉 🙌 Celebrate! Thanks for your contribution.

### DIFF
--- a/src/components/HTMLLoader.jsx
+++ b/src/components/HTMLLoader.jsx
@@ -34,7 +34,6 @@ function HTMLLoader({
     if (debouncedPostContent) {
       typeset(() => {
         if (previewRef.current !== null && typeof window?.MathJax !== 'undefined') {
-          window.MathJax.typesetClear();
           previewRef.current.innerHTML = sanitizedMath;
         }
       });


AhtishamShahid (commented):
what was the reason for calling this function and will it cause any inconsistency by removing it?
```js
MathJax.typesetClear();
```

awais-ansari (approved):

ayeshoali:
> what was the reason for calling this function and will it cause any inconsistency by removing it?
> 
> ```js
> MathJax.typesetClear();
> ```

MathJax.typesetClear() removes the typeset math from the page when new page is loaded.
but In version 3, there is information stored about the math in a list of typeset expressions, and when typeset math is removed from the page and replaced with new math, that list will hold pointers to math that no longer exists on the page which was causing this error.
It will not cause any inconsistency.

AhtishamShahid (approved):

PR #614, opened 2023-11-29
fix: resolved load more posts delay issue (quince.master)

awais-ansari:
### **[INF-1090](https://2u-internal.atlassian.net/browse/INF-1090)**

**Description** 
I noticed in Hotjar recordings that the “Load more posts” button is very slow and gets even slower with successive clicks. I’ve attached a recording of a test I conducted.
- First, load more clicks: API response takes 983 msec but it takes 6 seconds to load 10 new posts in the UI. :open_mouth:  
- 12th load more click: API response takes 816 msec but it takes 42 seconds to load 10 new posts in the UI. :exploding_head: It seems like the API call is sent 14 seconds after the button clicks.
https://www.loom.com/share/c2e296f135734a6389a238768d6e4602?sid=d0c0d7cb-7e9c-4cc5-b519-99c3c31bf9ad

**Implementation** 
- Component is optimized and every Post item is rendered one time. 
- The post item component was using the Truncate component to truncate the post title and preview. Which was taking too much time. 
- Our use cases are simple. So truncated the post title and preview using Bootstrap classes.

Screenshot 2023-11-06 at 9 55 07 PM

https://www.loom.com/share/4453d72e480243888de2a89cf206ce8f?sid=4f0da169-c032-4f40-99f5-ae6800f6b7aa

### DIFF
--- a/src/discussions/posts/post/PostLink.jsx
+++ b/src/discussions/posts/post/PostLink.jsx
@@ -1,4 +1,3 @@
-/* eslint-disable react/no-unknown-property */
 import React, { useContext, useMemo } from 'react';
 import PropTypes from 'prop-types';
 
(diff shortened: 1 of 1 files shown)


sundasnoreen12 (approved):
PR LGTM

PR #638, opened 2024-01-06
feat: removed sort and filter tour

awais-ansari:
[INF-1172](https://2u-internal.atlassian.net/browse/INF-1172)

### Description
We have 2 product tours currently active:
- Unresponded filter in discussions
- Responses/comments sort in discussions

These features are not “new” especially to ~2000 users signing up daily. They have 3 more clicks to do before they can interact with discussions and the product tours may be somewhat distracting to a user. We need to deactivate all product tours that we created.


sundasnoreen12 (approved):
PR LGTM

ayesha-waris (approved):

PR #659, opened 2024-01-31
fix: fixed redirection to learners tab in inContext view

ayesha-waris:
[INF-1224](https://2u-internal.atlassian.net/browse/INF-1224)

**Description**

Fixed the issue of redirection to the learners tab within the inContext view when username is clicked from post details.
The username in the inContext view is now simple text.

**Before

https://github.com/openedx/frontend-app-discussions/assets/73840786/90faf5a5-2e2b-416c-b2a6-22ddb8232955

After**

https://github.com/openedx/frontend-app-discussions/assets/73840786/c8c98d0a-ac86-4f15-9a04-f2def9229ac3


src/discussions/common/AuthorLabel.test.jsx
```diff
@@ -90,6 +90,15 @@ describe('Author label', () => {
...
 
+    it(
+      'it is not clickable when enableInContextSidebar is true',
+      async () => {
+        renderComponent(author, authorLabel, linkToProfile, labelColor, true);
+
+        expect(screen.queryByTestId('learner-posts-link')).not.toBeInTheDocument();
+      },
+    );
+
```
sundasnoreen12, line 101: Combine these two test cases using test.each and provide more appropriate descriptions for better clarity and readability

sundasnoreen12 (changes requested):

src/discussions/common/AuthorLabel.test.jsx
ayesha-waris: Both these test cases are testing two different scenarios. can not be coming with test.each

awais-ansari (approved):

sundasnoreen12 (approved):

Issue #688, opened 2024-04-02
It's possible to vote for your own posts in Discussions MFE


DmytroAlipov:
The user has the opportunity to vote for his post:
<img width="1284" alt="screen_67" src="https://github.com/openedx/frontend-app-discussions/assets/98233552/9eafb617-7e1c-48c5-9cfd-afba31e383a4">

But if you look at how it worked for Discussion XBlock, you can see that the ability to vote for your post is blocked and the button is not active for the user:
<img width="1284" alt="screen_68" src="https://github.com/openedx/frontend-app-discussions/assets/98233552/172da4ed-fad8-4eb1-9dc0-9aede2b69196">

I found in the documentation that the user should not be able to vote for his post [Discussions Docs](https://edx.readthedocs.io/projects/open-edx-building-and-running-a-course/en/latest/manage_discussions/running_discussions.html#minimizing-thread-proliferation:~:text=Vote%20for%20posts%20or%20responses.%20Learners%20can%20sort%20discussions%20by%20posts%20with%20the%20most%20votes%2C%20so%20posts%20and%20responses%20with%20many%20votes%20are%20more%20likely%20to%20be%20read%20and%20responded%20to.%20Select%20the%20%E2%80%9Cplus%E2%80%9D%20icon%20for%20the%20response.%20You%20cannot%20vote%20for%20your%20own%20posts.)

I found out that this resolution is obtained on the [backend](https://github.com/openedx/edx-platform/blob/7ad225658f3782f3e7a65e66702ab2898463bbc0/lms/djangoapps/discussion/django_comment_client/utils.py#L609). But this is not used to Discussion MFE.

It might be worth paying attention to this.

DmytroAlipov:
Hi @jmakowski1123.
Perhaps you can suggest who works more closely with this MFE to consult on this issue.
thanks in advance.

DmytroAlipov:
Hi, @arbrandes.
Are you involved in this type of MFE? It's just that this bug can be fixed in different ways. Ideally, this should be done on the frontend. By analogy with Legacy. This is much more difficult to do on the backend. Since all paths lead to the Forum service. I need an expert opinion before fixing this bug. I would appreciate your advice.

ayesha-waris:
@DmytroAlipov Thanks for pointing out this bug, it will be resolved in [this ticket](https://2u-internal.atlassian.net/browse/INF-1446)

DmytroAlipov:
[...]

PR #720, opened 2024-06-25
fix: fixed author liking its own post

ayesha-waris:
[INF-1446](https://2u-internal.atlassian.net/browse/INF-1446)
**Description**
fix for https://github.com/openedx/frontend-app-discussions/issues/688

To prevent learners from accidentally voting on their own posts (previously possible), this frontend PR disables the like button for a learner's own posts.
[BE PR](https://github.com/openedx/edx-platform/pull/34993)


awais-ansari (approved):

AhtishamShahid (approved):

src/discussions/common/HoverCard.jsx
```diff
@@ -12,7 +12,7 @@ import classNames from 'classnames';
 import { useIntl } from '@edx/frontend-platform/i18n';
 
 import { ThreadType } from '../../data/constants';
-import { useUserPostingEnabled } from '../data/hooks';
+import { useHasLikePermission, useUserPostingEnabled } from '../data/hooks';
```
sundasnoreen12, line 15: i think the hook name should be useUserLikePermission or useUserHasLikePermission

src/discussions/common/HoverCard.jsx
```diff
@@ -33,6 +33,7 @@ const HoverCard = ({
   const { enableInContextSidebar } = useContext(DiscussionContext);
   const { isClosed } = useContext(PostCommentsContext);
   const isUserPrivilegedInPostingRestriction = useUserPostingEnabled();
+  const hasLikePermission = useHasLikePermission(contentType, id);
```
sundasnoreen12, line 36: instead of hasLikePermission we can use isUserHasLikePermission

sundasnoreen12 (changes requested):

src/discussions/common/HoverCard.jsx
ayesha-waris: I believe both naming convention represent same purpose hasLikePermission also works well gere

sundasnoreen12 (approved):

PR #756, opened 2025-03-13
Made icons inside topic module tab accessible

zubairshakoorarbisoft:
### Description

On Discussion page, Under topic tab, modules sub sections are not keyboard accessible. User was not able to move inside the module to access message thread count etc now user can go inside and press enter to open the details.

Ticket: https://tree.taiga.io/project/zaraahmed-tutor-indigo-accessibility/us/33

#### How Has This Been Tested?

Please describe in detail how you tested your changes.

#### Screenshots/sandbox (optional):
Include a link to the sandbox for design changes or screenshot for before and after. **Remove this section if it's not applicable.**

|Before|After|
|-------|-----|
|      |      |

#### Merge Checklist

* [ ] If your update includes visual changes, have they been reviewed by a designer? Send them a link to the Sandbox, if applicable.
* [ ] Is there adequate test coverage for your changes?

#### Post-merge Checklist

* [ ] Deploy the changes to prod after verifying on stage or ask **@openedx/edx-infinity** to do it. 
* [ ] 🎉 🙌 Celebrate! Thanks for your contribution.

### DIFF
--- a/src/components/TopicStats.jsx
+++ b/src/components/TopicStats.jsx
@@ -39,7 +39,7 @@ const TopicStats = ({
           </Tooltip>
         )}
       >
-        <div className="d-flex align-items-center mr-3.5">
+        <div className="d-flex align-items-center mr-3.5" tabIndex={0}>
           <Icon src={PostOutline} className="icon-size mr-2" />
           {threadCounts?.discussion || 0}
         </div>
@@ -57,7 +57,7 @@ const TopicStats = ({
           </Tooltip>
         )}
       >
-        <div className="d-flex align-items-center mr-3.5">
+        <div className="d-flex align-items-center mr-3.5" tabIndex={0}>
           <Icon src={HelpOutline} className="icon-size mr-2" />
           {threadCounts?.question || 0}
         </div>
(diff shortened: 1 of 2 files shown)


wittjeff:
There are multiple issues here. 
- The topics are clickable but don't take keyboard focus (as they do for the parent items). 
- The topics act as links or as tabs depending on what you do decide to do with focus when they're clicked (I recommend link semantics, and jump focus to the top of the associated discussion thread. 
- You are giving focus to a thing that apparently only shows a tooltip. I think it would be better to give the icons accessible names (equivalent to the text in the tooltip).
- - If the comment and question indicators take focus or accept clicks, they should be buttons or links.
- There are (now with this change) focusable things on top of focusable things, which will throw a warning with automated accessibility checkers.  This is WCAG conformant if the click targets are minimum 24x24px in size (I'm not sure if they are; usually there is no strict target size, but a strict requirement for 24x24px of clearance between click/touch targets. when they're overlapping that logically becomes a minimum target size.)
- SO MANY TAB STOPS.  This UI problem occurs in some other commonly used scenarios like Facebook, etc.  This got me stuck when I was making UI recommendations before. I'm sorry!  
-- Allow the user to move focus between list items in the discussion tree visa arrow keys (as with tree navigation).  This skips over the repeated icons. But you have implemented the needed arrow key focus change behavior in the parent list, but not here.??

Issue #753, opened 2025-02-14
React 18 Upgrade


brian-smith-tcril:
Part of https://github.com/openedx/platform-roadmap/issues/429

```[tasklist]
- [ ] https://github.com/openedx/frontend-app-discussions/pull/758
- [ ] https://github.com/openedx/frontend-app-discussions/pull/759
```

PR #758, opened 2025-03-25
chore(deps): update `@openedx` dependencies to versions that support React 18

brian-smith-tcril:
Part 1 of https://github.com/openedx/frontend-app-discussions/issues/753


arbrandes (approved):
Looks good!  I see we're awaiting `userEvents` here already. :)

PR #772, opened 2025-05-20
fix: fixed Tabs menu is not displaying the "more" button

bydawen:


e0d:
@bydawen can you have a look at the coverage issues?

